PR #3, opened 2016-05-24
Implement setting app specific environment vars

mvo5:
This implements setting environment variables that got defined via
the snapd daemon. This is useful to avoid having to manually write
wrappers in snapcraft. Instead the environment is part of meta/snap.yaml
and will get generated automatically on install by snapd.

Note that this does not yet work for LD_LIBRARY_PATH, the strong suspicion is that the apparmor profile in debian/usr.bin.ubuntu-core-launcher prevents that via some apparmor environment scrubbing functionality. When the debian/usr.bin.ubuntu-core-launcher profile is removed adding LD_LIBRARY_PATH works just fine.

### DIFF
--- a/debian/usr.bin.ubuntu-core-launcher
+++ b/debian/usr.bin.ubuntu-core-launcher
@@ -70,6 +70,9 @@
     # reading seccomp filters
     /var/lib/snapd/seccomp/profiles/* r,
 
+    # reading the environment
+    /var/lib/snapd/environment/* r,
+
     # set up snap-specific private /tmp dir
     capability chown,
     /tmp/ w,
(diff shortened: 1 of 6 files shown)


debian/usr.bin.ubuntu-core-launcher
```diff
@@ -70,6 +70,9 @@
     # reading seccomp filters
     /var/lib/snapd/seccomp/profiles/* r,
 
+    # reading the environment
```
zyga: This has to move to to usr.bin.snap-run now

tests/common.sh
```diff
@@ -6,8 +6,12 @@ TMP="$(mktemp -d)"
 trap "rm -rf $TMP" EXIT
 
 export SNAPPY_LAUNCHER_SECCOMP_PROFILE_DIR="$TMP"
+export SNAPPY_LAUNCHER_ENVIRONMENT_DIR="$TMP/environment"
```
zyga, line 10: I think we should stop doing this and just define SNAPPY_LAUNCHER_something_something_ROOT and have everything else be relative from there. We're getting mount profiles next (another directory, another variable).

mvo5:
Closing as gustavo has some clever ideas that require significant rework.

PR #5, opened 2016-05-24
Discard the useless duplicate argument

zyga:
Now snap-run is invoked with 'snap-name $SECURITY_TAG $COMMAND...'.
Previously, security tag was duplicated. In the future this will change
once again, so that security tag is derived from an argument containing
$SNAP_NAME and $APP_NAME.

### DIFF
--- a/compat/ubuntu-core-launcher
+++ b/compat/ubuntu-core-launcher
@@ -1,2 +1,3 @@
 #!/bin/sh
+shift
 exec /usr/bin/snap-run "$@"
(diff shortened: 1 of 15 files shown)


mvo5:
Does the compat/ubuntu-core-launcher needs an update for this change?

zyga:
Ah, yes, I didn't do it as my local branch switched to snap-run already. Good catch, thanks

compat/ubuntu-core-launcher
```diff
@@ -1,2 +1,3 @@
 #!/bin/sh
+shift
```
mvo5, line 2: Heh, that was simple :)

mvo5:
:+1:

PR #6, opened 2016-05-24
Clean up tests so that shellcheck reports no more errors or warnings

zyga:


tests/common.sh
```diff
@@ -1,12 +1,14 @@
...
 L="$(pwd)/../src/snap-run"
+export L
 
 TMP="$(mktemp -d)"
-trap "rm -rf $TMP" EXIT
+trap 'rm -rf $TMP' EXIT
 
 export SNAPPY_LAUNCHER_SECCOMP_PROFILE_DIR="$TMP"
 export SNAPPY_LAUNCHER_INSIDE_TESTS="1"
+export UBUNTU_CORE_LAUNCHER_NO_ROOT=1
```
zyga, line 11: This was supposed to be moved from run-tests so that each test was runnable standalone.

debian/changelog
```diff
@@ -9,6 +9,7 @@ ubuntu-core-launcher (2.0) UNRELEASED; urgency=medium
     'snap-name $SECURITY_TAG $COMMAND...'. Previously, security tag was
     duplicated. In the future this will change once again, so that security
     tag is derived from an argument containing $SNAP_NAME and $APP_NAME. 
+  * Clean up tests so that shellcheck reports no more errors or warnings
```
mvo5, line 12: I wonder if we should make `shellshock^Wshellcheck` part of  `make check` ? To avoid future regressions.

tests/common.sh
```diff
@@ -1,12 +1,14 @@
 #!/bin/sh
     
 L="$(pwd)/../src/snap-run"
+export L
 
 TMP="$(mktemp -d)"
-trap "rm -rf $TMP" EXIT
+trap 'rm -rf $TMP' EXIT
```
mvo5, line 7: Nice catch!

tests/test_bad_seccomp_filter_missing_trailing_newline
```diff
@@ -2,13 +2,13 @@
 
 set -e
 
-. $(pwd)/common.sh
+. "$(pwd)/common.sh"
 
 printf "Test seccomp filter (bad - no trailing newline)"
 
-echo -n "missingnewline" > $TMP/snap.name.app
+printf "missingnewline" > "$TMP/snap.name.app"
```
mvo5, line 9: Heh, shellcheck is a bit of a pedant. But yeah, it does make sense!

mvo5:
Very nice, many thanks. Shell is hard, I keep telling that everybody who wants to listen.

debian/changelog
zyga, line 12: Mmm, yeah. Good idea!

zyga, line 12: Done

mvo5:
:+1:

PR #8, opened 2016-05-25
allow to run /usr/lib/snapd/snap-exec

mvo5:
Prepare for moving to `snap run`. Note that this is not good enough yet, I get:

```
/usr/bin/snap: error while loading shared libraries: cannot apply additional memory protection after relocation: Permission denied
```

when trying to run /usr/bin/snap from within the launcher confinement.

The idea is that the wrapper will look like this:
`exec snap run command`

Then `snap run` will  do:
-  all the things it needs to do without confinement (get revision, setup SNAP_USER_DATA_DIR, potentially ask the daemon to refresf confinment etc).
- exec `ubuntu-core-launcher` (snap-run) with `snap run --after-suid-setup`
- snap run --after-suid-setup will parse the snap yaml (confined), will setup the environment including LD_LIBRARY_PATH and all the environment that we currently setup in the launcher
- exec the actual binary.

### DIFF
--- a/debian/usr.bin.ubuntu-core-launcher
+++ b/debian/usr.bin.ubuntu-core-launcher
@@ -19,6 +19,9 @@
 
     /usr/bin/ubuntu-core-launcher r,
 
+    # for snap run
+    /usr/bin/snap r,
+
     /dev/null rw,
     /dev/full rw,
     /dev/zero rw,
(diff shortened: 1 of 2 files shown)


debian/usr.bin.ubuntu-core-launcher
```diff
@@ -19,6 +19,10 @@
 
     /usr/bin/ubuntu-core-launcher r,
 
+    # for snap run
+    /usr/bin/snap mr,
+    /lib/@{multiarch}/linux-vdso.so* mr,
```
zyga: vdso is not a real shared library, the kernel maps and passes the location via the auxiliary vector. Are you sure this is required?

zyga:
Based on what we discussed I don't think this has to land. After at that time the executed executable (snap-exec) will be a part of the OS snap and that's white-listed already. Let's reopen this when we know more.

PR #10, opened 2016-05-27
Change how core snap mounts are created

zyga:
This patch fundamentally changes how core snap bind mounts are created
when running on a classic distribution. To align with how an all-snap
distribution looks like, the root filesystem is now a read only core
snap. From that snap a number of bind mounts reach to the classic system
(for user and system writable data and a few other things).

Lastly since the runtime environment is now identical in all places, we
should reset PATH to always aligned with the core snap.

### DIFF
--- a/debian/usr.bin.snap-run
+++ b/debian/usr.bin.snap-run
@@ -68,29 +68,38 @@
     # change_profile -> **,
 
     # reading seccomp filters
-    /var/lib/snapd/seccomp/profiles/* r,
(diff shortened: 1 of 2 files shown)


niemeyer:
This should really not be done now. Way too much is being changed at once before anything at all is released. Let's please focus on getting the snap run and snap-confine story out of the door before even more fundamental assumptions change.

zyga:
This is now merged as a part of the 1.0.29 release, closing.

PR #11, opened 2016-05-27
Fix maintainer script not to fail if profile is not loaded

zyga:

### DIFF
--- a/debian/ubuntu-core-launcher.postinst
+++ b/debian/ubuntu-core-launcher.postinst
@@ -8,7 +8,9 @@ case $1 in
     configure)
         # remove current profile so that the pass-through can still run
         if [ -e /sys/kernel/security/apparmor/.remove ]; then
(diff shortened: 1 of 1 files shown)


debian/ubuntu-core-launcher.postinst
```diff
@@ -8,7 +8,9 @@ case $1 in
     configure)
         # remove current profile so that the pass-through can still run
         if [ -e /sys/kernel/security/apparmor/.remove ]; then
-            echo -n '/usr/bin/ubuntu-core-launcher' > /sys/kernel/security/apparmor/.remove
+            if cat /sys/kernel/security/apparmor/profiles | grep /usr/bin/ubuntu-core-launcher; then
```
tyhicks: grep can read files so I'd suggest:

grep -q "^/usr/bin/ubuntu-core-launcher ([[:alpha:]]+)$" /sys/kernel/security/apparmor/profiles

zyga: Done

mvo5:
:+1:

PR #12, opened 2016-05-30
Fix prototype for seccomp_load_filters

zyga:
This patch fixes previously unnoticed discrepancy between the prototype and definition of `seccomp_load_ilters()` 


mvo5:
Hrm, `-Wall` and all that. C is just hard. Thanks for finding this.

PR #14, opened 2016-05-30
Use autotools build system

zyga:
This patch discards the hand-crafted build system in favour of autotools.
This change makes it easier to build the source code and to begin adding
conditional features.

The seccomp.[ch] file was renamed to seccomp_utils.[ch] in order not to
clash with the system-wide seccomp.h. This is required because of how
autotools pass the -I. flag by default.

Unit tests were converted to autotool standards.


src/Makefile.am
```diff
@@ -0,0 +1,11 @@
+bin_PROGRAMS = snap-run
+snap_run_SOURCES = main.c seccomp_utils.c utils.c
+snap_run_CFLAGS = \
```
mvo5, line 3: Hm, no more `make fmt` and `make syntax-check` ? And will `make test` work as before? It is also dropping the udev releated stuff unless I miss something.

zyga, line 3: `make fmt` is indeed gone, I'll bring it back. Make test is the autotools-std `make check` instead.
I see I don't install the udev helper, thanks for spotting that!

zyga, line 3: check syntax restored (for shell scripts), looking at .c and installing the small udev helper

zyga, line 3: check syntax restored (for C files), looking at the last thing

zyga, line 3: Udev rules restored, testing with local packages now

zyga:
make fmt restored

configure.ac
```diff
@@ -0,0 +1,41 @@
...
+AC_TYPE_MODE_T
+AC_TYPE_PID_T
+AC_TYPE_SIZE_T
+
+# Checks for library functions.
+AC_FUNC_CHOWN
+AC_FUNC_ERROR_AT_LINE
+AC_FUNC_FORK
+AC_FUNC_STRNLEN
+AC_CHECK_FUNCS([mkdir regcomp setenv strdup strerror])
```
tyhicks, line 33: I don't know if it is important to check all the functions used by snap-confine here (I typically don't do it in the projects that I'm involved in) but there are some missing functions in this list.

tyhicks:
Have you done a comparison of the build logs before and after applying this PR to verify that compiler flags did not change for the worse?

configure.ac
zyga, line 33: This is just autoscan, I suspect that we can just remove this whole chunk as those functions don't really differ across systems anymore.

Thanks for having a look!

zyga:
@tyhicks I didn't do that, that's a good idea.

debian/control
```diff
@@ -2,7 +2,7 @@ Source: ubuntu-core-launcher
 Section: utils
 Priority: optional
 Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
-Build-Depends: debhelper (>= 9), libseccomp-dev, libapparmor-dev, libudev-dev, dh-apparmor, indent, shellcheck
+Build-Depends: debhelper (>= 9), libseccomp-dev, libapparmor-dev, libudev-dev, dh-apparmor, indent, shellcheck, pkg-config
```
ogra1: don't you need autotools-dev and autoconf, automake build deps as well here ?

zyga: Not according to running this in sbuild. I can add it though.

tyhicks:
I tried to build a debian package with sbuild and it fails:

  https://paste.ubuntu.com/16897029/

Nothing is being built. This diff between builds before and after this PR is applied shows that to be the case:

  https://paste.ubuntu.com/16897178/

zyga:
Thanks for testing this. Let me try to reproduce that.

zyga:
Ah, indeed, I just had a dirty tree with ./configure around. Fixed now.

debian/control
zyga: Now with --autoreconf everything works. Thanks for correcting me ogra!

zyga:
Merging based on review by @tyhicks and ogra on IRC

PR #18, opened 2016-06-01
Add --disable-confinement configure option

zyga:
This patch adds a configure option to disable strict confinement. With this
option selected apparmor is no longer a dependency and snap-run can be used
anywhere, regardless of the kernel patches or availability of apparmor, as long
snaps are used in devmode.

Signed-off-by: Zygmunt Krynicki zkrynicki@gmail.com

### DIFF
--- a/PORTING
+++ b/PORTING
@@ -1,0 +1,15 @@
+Welcome brave porters!
+
+This file is intended to guide you towards porting snappy (comprised of snapd
+and this project, snap-confine) to work on a new kernel. The confinement setup by
+snap-confine has several requirements on the kernel.
+
+TODO: list required patches (apparmor, seccomp)
+TODO: list required kernel configufation
+TODO: list minimum supported kernel version
+
+While you are working on porting those patches to your kernel of choice, you
+may configure snap-confine with --disable-security. This switch drops
+requirement on apparmor, seccomp and udev and reduces snap-confine to arrange
+the filesystem in a correct way for snaps to operate without really confining
+them in any way.
--- a/src/main.c
+++ b/src/main.c
@@ -28,7 +28,7 @@
 #include <sys/mount.h>
 #ifdef STRICT_CONFINEMENT
 #include <sys/apparmor.h>
-#endif
+#endif				// ifdef STRICT_CONFINEMENT
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
@@ -43,13 +43,15 @@
 
 #include <ctype.h>
 
-#include "libudev.h"
-
 #include "utils.h"
+#ifdef STRICT_CONFINEMENT
+#include "libudev.h"
 #include "seccomp_utils.h"
+#endif				// ifdef STRICT_CONFINEMENT
 
 #define MAX_BUF 1000
 
+#ifdef STRICT_CONFINEMENT
 struct snappy_udev {
 	struct udev *udev;
 	struct udev_enumerate *devices;
@@ -57,6 +59,7 @@ struct snappy_udev {
 	char tagname[MAX_BUF];
 	size_t tagname_len;
 };
+#endif				// ifdef STRICT_CONFINEMENT
 
 bool verify_appname(const char *appname)
 {
@@ -76,6 +79,7 @@ bool verify_appname(const char *appname)
 	return (status == 0);
 }
 
+#ifdef STRICT_CONFINEMENT
 void run_snappy_app_dev_add(struct snappy_udev *udev_s, const char *path)
 {
 	if (udev_s == NULL)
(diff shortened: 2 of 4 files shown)


src/main.c
```diff
@@ -545,13 +549,15 @@ int main(int argc, char **argv)
...
 	int rc = 0;
 	// set apparmor rules
 	rc = aa_change_onexec(aa_profile);
 	if (rc != 0) {
 		if (secure_getenv("SNAPPY_LAUNCHER_INSIDE_TESTS") == NULL)
 			die("aa_change_onexec failed with %i", rc);
 	}
+#endif
 	// set seccomp (note: seccomp_load_filters die()s on all failures)
 	seccomp_load_filters(aa_profile);
```
jdstrand, line 573: Why aren't the other sandboxing features disabled as well? For example, seccomp needs a particular kernel and very soon will need a newer kernel/particular patches backported to work with snapd designs. '--disable-confinement' suggests to me disabling much more than apparmor. It should disable apparmor, seccomp, devpts, udev/cgroups and private mount. I would imagine that with --disable-confinement you would want to also strip the setuid bit as well. I'm not sure how useful it would be for people bootstrapping the confinement story, but perhaps in a future commit you could allow things like:
- --disable-confinement-apparmor # disables apparmor
- --disable-confinement-cgroups # disables cgroups
- --disable-confinement-devpts # disables devpts
- --disable-confinement-private-tmp # disables private tmp
- --disable-confinement-seccomp # disables seccomp
- --disable-confinement # disables everything in one go

Please don't block on --disable-confinement-\* though.

zyga:
Thanks for commenting @jdstrand. I have a much better picture as to what features depend on the kernel more strongly than others. I can work on splitting the confinement as suggested (though in another branch, today I will just put apparmor and secomp into one bag).

Do you think we could maintain a PORTING file in the tree so that it is clear which kernel features / configuration flags are required. I would love to know how porters can track those and know when it is sensible/safe to drop one of the `--disable-confinement-foo` switches.

I know we have some good documentation for that elsewhere on the wiki but perhaps linking it here would be sufficient.

tyhicks:
We had previously decided that confinement is either enabled or disabled (devmode). We had decided that if we present confinement as enabled, a degraded/incomplete level of confinement would be misleading the user since snaps could break out of confinement.

We were specifically talking about the strength the of AppArmor confinement at that time but I think it extends equally well to the other pieces of the confinement story. Disabling one piece of confinement, while enabling others, puts us in a state with a degraded/incomplete level of confinement.

jdstrand:
@zyga - Leann is collecting and maintaining a list of required patches for the different components (for sandboxing, kernel configs, systemd, etc). I'm not sure of the messaging surrounding that or the status of a proper porting guide but I agree that all of this should be documented. I would suggest when that information is in its final form that a PORTING file should point to it.

jdstrand:
@tyhicks - I agree that for target systems it should be all or nothing but I'm less sure about how to support porters-- it seems that giving them the ability to pick and choose would more easily allow them to fix/test the pieces as they develop them, but you're right there is no way to control what they will do with those options.

I wonder if there is another way to handle this-- perhaps add a '--enable-devmode' option that then allows the launcher to honor environment or command line options? A bootstrapper would compile with '--enable-devmode' and then snap-confine could say, examine SNAP_CONFINE_APPARMOR, SNAP_CONFINE_SECCOMP, SNAP_CONFINE_DEVPTS, etc. A fully-bootstrapped system would not be compiled with '--enable-devmode' and all those env vars would be ignored.

UPDATE: note, I am suggesting adding --enable-devmode in place of --disable-confinement

tyhicks:
@jdstrand I think we're making this more complex than it needs to be. I just don't see porters enabling and pulling in patches for AppArmor, building a kernel to test, then enabling and pulling in patches for seccomp, then building a new kernel to test, etc. I think they're fine to pull all the needed patches in, build one kernel, and test that kernel.

A setuid root binary examining environment variables to decide on whether or not to enable confinement also sounds iffy to me. We should be using secure_getenv() to check environment variables and that's going to return NULL when the setuid bit is set on the binary, anyways.

jdstrand:
It's certainly easier to simply disable everything.

@zyga, perhaps disable all of: apparmor, seccomp, cgroups, devpts and private tmp when compiled with '--disable-confinement'. If selective disabling becomes a requirement, we can revisit.

zyga:
@tyhicks @jdstrand I'll go ahead with fully disabled everything

PORTING
```diff
@@ -0,0 +1,15 @@
...
+and this project, snap-confine) to work on a new kernel. The confinement setup by
+snap-confine has several requirements on the kernel.
+
+TODO: list required patches (apparmor, seccomp)
+TODO: list required kernel configufation
+TODO: list minimum supported kernel version
+
+While you are working on porting those patches to your kernel of choice, you
+may configure snap-confine with --disable-security. This switch drops
+requirement on apparmor, seccomp and udev and reduces snap-confine to arrange
```
jdstrand, line 13: Minot nit: the path actually drops apparmor, seccomp, udev, private mount and devpts newinstance.

PORTING
```diff
@@ -0,0 +1,15 @@
+Welcome brave porters!
+
+This file is intended to guide you towards porting snappy (comprised of snapd
+and this project, snap-confine) to work on a new kernel. The confinement setup by
+snap-confine has several requirements on the kernel.
+
+TODO: list required patches (apparmor, seccomp)
+TODO: list required kernel configufation
+TODO: list minimum supported kernel version
```
jdstrand, line 9: It looks like this is the link: https://developer.ubuntu.com/en/snappy/guides/porting/ but it still hasn't been updated yet.

PR #20, opened 2016-06-02
Skip confinment tests if confinement is disabled

zyga:

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1,4 +1,7 @@
-TESTS = \
+TESTS =
+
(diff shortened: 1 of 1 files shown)


jdstrand:
This is fine in general, but you definitely want to test something when confinement is disabled. Testing that something launches at all and that userdata is created come to mind. I suggest adding a test script or two under an 'else' to the 'if STRICT_CONFINEMENT'.

zyga:
@jdstrand that can be added in the first line as soon as we have that test. I'll add that to my TODO pile

PR #21, opened 2016-06-02
Move the lintian overrides to the new package name

vorlonofportland:
snap-run was renamed to snap-confine before the lintian overrides landed, move them now


mvo5:
:+1: Thank you!

PR #25, opened 2016-06-06
Build compatiblity script to not hardcode $(libdir)

zyga:
Since $(libdir) is configurable it should not be hard-coded as /usr/lib anywhere, including the compatibility script.

### DIFF
--- a/compat/Makefile.am
+++ b/compat/Makefile.am
@@ -1,1 +1,6 @@
 dist_bin_SCRIPTS = ubuntu-core-launcher
+CLEANFILES = $(dist_bin_SCRIPTS)
+
+ubuntu-core-launcher: ubuntu-core-launcher.in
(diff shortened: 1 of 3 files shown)


mvo5:
:+1:

compat/Makefile.am
```diff
@@ -1 +1,6 @@
 dist_bin_SCRIPTS = ubuntu-core-launcher
+CLEANFILES = $(dist_bin_SCRIPTS)
+
+ubuntu-core-launcher: ubuntu-core-launcher.in
+	sed -e 's,[@]LIBDIR[@],$(libdir),g' < $(srcdir)/ubuntu-core-launcher.in > ubuntu-core-launcher
```
mvo5, line 5: A bit sad that this can not be done in a more elegant way

zyga, line 5: On the up side this script is going away soon :-)

PR #33, opened 2016-06-07
Implement --enable-rootfs-is-core-snap

zyga:
This branch implements the configuration option to use the core snap as the root file system.

### DIFF
--- a/debian/usr.bin.snap-confine
+++ b/debian/usr.bin.snap-confine
@@ -68,29 +68,38 @@
     # change_profile -> **,
 
     # reading seccomp filters
-    /var/lib/snapd/seccomp/profiles/* r,
+    /{tmp/snap.rootfs_*/,}var/lib/snapd/seccomp/profiles/* r,
 
     # set up snap-specific private /tmp dir
     capability chown,
-    /tmp/ w,
-    /tmp/snap.*/ w,
-    /tmp/snap.*/tmp/ w,
-    mount options=(rw private) -> /tmp/,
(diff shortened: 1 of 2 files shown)


debian/usr.bin.snap-confine
```diff
@@ -68,29 +68,38 @@
...
-    mount options=(rw bind) /snap/ubuntu-core/*/bin/ -> /bin/,
-    mount options=(rw bind) /snap/ubuntu-core/*/sbin/ -> /sbin/,
-    mount options=(rw bind) /snap/ubuntu-core/*/lib/ -> /lib/,
-    mount options=(rw bind) /snap/ubuntu-core/*/lib32/ -> /lib32/,
-    mount options=(rw bind) /snap/ubuntu-core/*/libx32/ -> /libx32/,
-    mount options=(rw bind) /snap/ubuntu-core/*/lib64/ -> /lib64/,
-    mount options=(rw bind) /snap/ubuntu-core/*/usr/ -> /usr/,
+    /{tmp/snap.rootfs_*,}snap/ r,
+    /{tmp/snap.rootfs_*,}snap/** r,
+    mount options=(rw bind) /snap/ubuntu-core/*/ -> /tmp/snap.rootfs_*/,
```
jdstrand, line 87: Can you add a comment for this and why it is different than the next group or rules (you can add a comment there too).

zyga, line 87: Ah, good catch. I think I should have kept the original set of options

zyga:
I'll close this as all of the branches are landing through the new-release branch.

PR #37, opened 2016-06-14
Add a few more libraries that the 361 nvidia driver needs

tseliot:
This simply adds a few more libraries that the 361 driver needs.

Krita still segfaults with this though, and the segfault points to glibc.


zyga:
Based on feedback provided via other channels I'm merging this. This doesn't make krita work but it does make the notes application work and that's an improvement. We'll keep exploring why krita is failing.

PR #38, opened 2016-06-15
chdir() to '/' before setting up private /tmp (LP: #1592402)

jdstrand:
chdir() to '/' before setting up private /tmp so private /tmp works when user is in /tmp (LP: #1592402)


mvo5:
:+1: Would be nice to mention in the comment that this is an issue if the user is in /tmp (that is my understanding at least)

zyga:
This looks good. Thank you Jamie.

One word of notice: we're trying to unify error messages around `cannot ...` rather than `unable to`. I will merge it and update the tree to reflect that.

zyga:
I'll merge this manually in a sec.

zyga:
Thanks for that and for de-conflicting. Merging now

PR #40, opened 2016-06-16
single binary, even for wrapper

teknoraver:
remove compatibility wrapper

### DIFF
--- a/compat/Makefile.am
+++ b/compat/Makefile.am
@@ -1,7 +1,5 @@
 dist_bin_SCRIPTS = ubuntu-core-launcher
 CLEANFILES = $(dist_bin_SCRIPTS)
-EXTRA_DIST = ubuntu-core-launcher.in
 
-ubuntu-core-launcher: ubuntu-core-launcher.in Makefile
-	sed -e 's,[@]libexecdir[@],$(libexecdir),g' < $(srcdir)/ubuntu-core-launcher.in > ubuntu-core-launcher
-	chmod +x ubuntu-core-launcher
+ubuntu-core-launcher: Makefile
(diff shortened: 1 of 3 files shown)


zyga:
This does look nice, thank you. Let me do some testing (we don't have any automated CI setup for this repository)

teknoraver:
The argument shift seems to work:

```
root@ubuntu32:~/src/snap-confine# src/snap-confine arg1 arg2
appname arg1 not allowed
root@ubuntu32:~/src/snap-confine# compat/ubuntu-core-launcher arg_old arg1 arg2
appname arg1 not allowed
```

compat/Makefile.am
```diff
@@ -1,7 +1,5 @@
 dist_bin_SCRIPTS = ubuntu-core-launcher
 CLEANFILES = $(dist_bin_SCRIPTS)
-EXTRA_DIST = ubuntu-core-launcher.in
 
-ubuntu-core-launcher: ubuntu-core-launcher.in Makefile
-	sed -e 's,[@]libexecdir[@],$(libexecdir),g' < $(srcdir)/ubuntu-core-launcher.in > ubuntu-core-launcher
-	chmod +x ubuntu-core-launcher
+ubuntu-core-launcher: Makefile
+	ln -s snap-confine ubuntu-core-launcher
```
zyga, line 5: Note that the location of snap-confine and ubuntu-core-launcher can be rather different. I will need to experiment with how this affects downstream distributions. Some packages (_ahem_ dpkg) don't deal very well with file/symlink transitions AFAIR

mvo5:
Symlinks to files should be fine with dpkg. Symlinks to directories are problematic because dpkg tries to be smart about them

PR #44, opened 2016-06-20
Constrain tests to x86_64/amd64

zyga:
This patch changes tests so that they are only ran on amd64. Tests are
somewhat syscall specific and on some architectures they cause false
alarms today. Before this is fixed they should be constrained.


mvo5:
:+1:

PR #49, opened 2016-06-23
Spreadify

zyga:


zyga:
Tests pass but be aware of the fact that we only have one `debian-8` node on linode and when too many tests run at the same time things tend to fail.

mvo5:
Nice!

PR #51, opened 2016-06-27
Add support for mount profiles

zyga:
This branch adds support for the new mount profiles. Snapd can now optionally create files in `/var/lib/snapd/mount/*.fstab`. Each file is associated with exactly one application or hook. The file has syntax that of `fstab(5)` and describes mounts that should be created in the private mount namespace of the started application or hook.

The set of actually allowed mounts is restricted, currently only bind mounts are supported.

This branch is still a work in progress, please don't review it yet.


src/cleanup-funcs.h
```diff
@@ -26,4 +28,20 @@
  **/
 void sc_cleanup_string(char **ptr);
 
+/**
+ * Close an open file.
+ *
+ * This function is designed to be used with
+ * __attribute__((cleanup(sc_cleanup_file))).
```
tyhicks, line 35: I'd suggest defining an easy-to-type macro for "**attribute** ((cleanup(sc_cleanup_endmntent)))". I think "autofclose" is descriptive.

zyga:
Merging now to iterate on safe mount and more unit tests

PR #52, opened 2016-06-28
Add a manual page for snap-confine

zyga:
While snap-confine is not typically exposed on PATH, having a manual
page will help to demystify some of the things happening inside.

I plan to land this small version an then expand it with more details
about how the execution environment looks like.

### DIFF
--- a/docs/snap-confine.rst
+++ b/docs/snap-confine.rst
@@ -1,0 +1,117 @@
+==============
+ snap-confine
+==============
+
+-----------------------------------------------
+internal tool for confining snappy applications
+-----------------------------------------------
+
+:Author: zygmunt.krynicki@canonical.com
+:Date:   2016-06-27
+:Copyright: Canonical Ltd.
+:Version: 1.0.33
+:Manual section: 5
+:Manual group: snappy
+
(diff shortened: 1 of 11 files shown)


docs/snap-confine.rst
```diff
@@ -0,0 +1,117 @@
...
+`snap-confine` responds to the following environment variables
+
+`UBUNTU_CORE_LAUNCHER_DEBUG`:
+	When defined the program will print additional diagnostic information about
+	the actions being performed. All the output goes to stderr.
+
+The following variables are only used when `snap-confine` is not setuid root.
+This is only applicable when testing the program itself.
+
+`SNAPPY_LAUNCHER_INSIDE_TESTS`:
```
mvo5, line 88: Maybe we should just not mention those? WIth the new (very nice) unittest work we may kill them in the future I guess?

mvo5:
:+1:

docs/snap-confine.rst
zyga, line 88: I've mentioned them for completeness. I do plan to remove them (and to rename remaining ubuntu  bits )

PR #56, opened 2016-06-30
Install udev rules even if confinement isn't enabled

mwhudson:
Even if confinement is not enabled in snapd, it's safe to install the udev rule
and the udev helper which simply will never match any tags.  Installing these
means we can avoid special-casing of the packaging on different Debian-based
distributions.

(this is the patch that is currently in debian sid)


zyga:
Hey @mwhudson, long time no see :-)

The patch makes sense, thanks.

PR #58, opened 2016-06-30
Split real main and test main

zyga:


zyga:
This is just something that mvo asked for earlier

PR #59, opened 2016-06-30
Replace remantns of ubuntu-core-launcher

zyga:
This branch replaces some remains of ubuntu-core-launcher names with snap-confine.


pete-woods:
Looks pretty reasonable to me 👍

PR #62, opened 2016-06-30
debian/usr.bin.snap-confine: allow read of /dev/random and /dev/urandom

jdstrand:


zyga:
+1, just in time Jamie :-)

jdstrand:
Heh, and here I thought I missed the window. Thanks! :)

PR #64, opened 2016-06-30
Add support for running hooks.

kyrofa:
This involves a move away from "appname" to "security tag".

### DIFF
--- a/src/sc-main.c
+++ b/src/sc-main.c
@@ -54,19 +54,15 @@ int sc_main(int argc, char **argv)
 	if (argc < NR_ARGS + 1)
 		die("Usage: %s <security-tag> <binary>", argv[0]);
 
-	const char *executable_name = argv[1];
-	debug("executable name is %s", executable_name);
-#ifdef STRICT_CONFINEMENT
-	const char *aa_profile = argv[1];
-	debug("security-tag is %s", aa_profile);
-#endif				// ifdef STRICT_CONFINEMENT
+	const char *security_tag = argv[1];
+	debug("security tag is %s", security_tag);
 	const char *binary = argv[2];
 	debug("binary to run is %s", binary);
 	uid_t real_uid = getuid();
 	gid_t real_gid = getgid();
 
-	if (!verify_executable_name(executable_name))
-		die("executable name %s not allowed", executable_name);
+	if (!verify_security_tag(security_tag))
+		die("security tag %s not allowed", security_tag);
 
 	// this code always needs to run as root for the cgroup/udev setup,
 	// however for the tests we allow it to run as non-root
@@ -103,20 +99,20 @@ int sc_main(int argc, char **argv)
 		}
 #ifdef STRICT_CONFINEMENT
 		// set up private mounts
-		setup_private_mount(executable_name);
+		setup_private_mount(security_tag);
 
 		// set up private /dev/pts
 		setup_private_pts();
 
 		// this needs to happen as root
 		struct snappy_udev udev_s;
-		if (snappy_udev_init(executable_name, &udev_s) == 0)
-			setup_devices_cgroup(executable_name, &udev_s);
+		if (snappy_udev_init(security_tag, &udev_s) == 0)
+			setup_devices_cgroup(security_tag, &udev_s);
 		snappy_udev_cleanup(&udev_s);
 #endif				// ifdef STRICT_CONFINEMENT
 
 		// setup the security backend bind mounts
-		sc_setup_mount_profiles(executable_name);
+		sc_setup_mount_profiles(security_tag);
 
 		// Try to re-locate back to vanilla working directory. This can fail
 		// because that directory is no longer present.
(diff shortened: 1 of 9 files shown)


src/sc-main.c
```diff
@@ -54,19 +54,19 @@ int sc_main(int argc, char **argv)
 	if (argc < NR_ARGS + 1)
 		die("Usage: %s <security-tag> <binary>", argv[0]);
 
-	const char *appname = argv[1];
-	debug("appname is %s", appname);
+	const char *executable_name = argv[1];
```
zyga: In reality this is `secuirty_tag`.

zyga: If you are changing this, can you just call it as such, and remove `aa_profile` entirely as it was always the same thing

src/snap.c
```diff
@@ -23,19 +23,22 @@
...
+	// snap.<name>.(<appname>|hook.<hookname>)
 	// - <name> must start with lowercase letter, then may contain
 	//   lowercase alphanumerics and '-'
-	// - <app> may contain alphanumerics and '-'
-	const char *whitelist_re = "^snap\\.[a-z][a-z0-9-]*\\.[a-zA-Z0-9-]+$";
+	// - <appname> may contain alphanumerics and '-'
+	// - <hookname must start with a lowercase letter, then may
+	//   contain lowercase letters and '-'
+	const char *whitelist_re =
+	    "^snap\\.[a-z](-?[a-z0-9])*\\.([a-zA-Z0-9](-?[a-zA-Z0-9])*|hook\\.[a-z](-?[a-z])*)$";
```
zyga, line 36: Can you perhaps link to the snapd code that does similar checks?

src/verify-executable-name-test.c
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+#include "snap.h"
+#include "snap.c"
+
+#include <glib.h>
+
+static void test_verify_executable_name()
+{
+	// First, test the names we know are good
```
zyga, line 25: Nice, thank you!

src/verify-executable-name-test.c
```diff
@@ -0,0 +1,67 @@
...
+	g_assert_true(verify_executable_name("snap.name.app"));
+	g_assert_true(verify_executable_name
+		      ("snap.network-manager.NetworkManager"));
+	g_assert_true(verify_executable_name("snap.f00.bar-baz1"));
+	g_assert_true(verify_executable_name("snap.foo.hook.bar"));
+	g_assert_true(verify_executable_name("snap.foo.hook.bar-baz"));
+
```
zyga, line 32: Can you add a canary test for `"snap.name"`. At some point we may decide to abbreviate the security tag and this would be a nice test to catch that change.

src/snap.c
```diff
@@ -23,19 +23,22 @@
...
-	// snap.<name>.<app>
+	// The executable name is of form:
+	// snap.<name>.(<appname>|hook.<hookname>)
 	// - <name> must start with lowercase letter, then may contain
 	//   lowercase alphanumerics and '-'
-	// - <app> may contain alphanumerics and '-'
-	const char *whitelist_re = "^snap\\.[a-z][a-z0-9-]*\\.[a-zA-Z0-9-]+$";
+	// - <appname> may contain alphanumerics and '-'
+	// - <hookname must start with a lowercase letter, then may
+	//   contain lowercase letters and '-'
```
zyga, line 34: can hook names contain digits?

src/snap.c
```diff
@@ -23,19 +23,22 @@
...
+	const char *whitelist_re =
+	    "^snap\\.[a-z](-?[a-z0-9])*\\.([a-zA-Z0-9](-?[a-zA-Z0-9])*|hook\\.[a-z](-?[a-z])*)$";
 	regex_t re;
 	if (regcomp(&re, whitelist_re, REG_EXTENDED | REG_NOSUB) != 0)
 		die("can not compile regex %s", whitelist_re);
 
-	int status = regexec(&re, appname, 0, NULL, 0);
```
zyga, line 38: Just thinking about something nasty being done. Can we require that total length of `security_tag` be smaller than some sane default. I don't want to let people explode the system by installing a snap with a long app name that is expensive to check.

If there is something in snapd, let's use that limit, if not then let's make up one now.

@jdstrand ^^ FYI

src/snap.c
kyrofa, line 34: No, just lowercase letters and hyphens.

src/snap.c
kyrofa, line 36: It's [here](https://github.com/snapcore/snapd/blob/master/snap/validate.go), are you saying you want that link in the code?

src/sc-main.c
kyrofa: Yeah might as well.

src/snap.c
kyrofa, line 38: I don't know of a limit imposed by snapd, so throw out a number! Though that may be something for another PR.

jdstrand, line 38: There are practical limits to how long this can be and the review tools enforce this. The max length enforced by the store is '100' for the full security label ('snap.$snap.$app').

src/sc-main.c
kyrofa: Done.

src/verify-executable-name-test.c
kyrofa, line 32: Done.

src/snap.c
kyrofa, line 38: @zyga do you want that in this PR, or should it be done separately?

src/snap.c
zyga, line 34: OK, thanks

src/snap.c
zyga, line 38: This can go in as there wasn't any check around there before. I can take care of that.

zyga:
+1, thanks for working on this

PR #65, opened 2016-06-30
Towards safe mount

zyga:
Just more functions and more tests

### DIFF
--- a/src/mount-support.c
+++ b/src/mount-support.c
@@ -393,22 +393,34 @@ static char * __attribute__ ((used))
 	return (offset < fulllen) ? &path[offset] : NULL;
 }
 
-/*
+/**
  * Check that @subdir is a subdir of @dir.
- * @len is the length of @dir (to avoid having to recalculate it).
 **/
 static bool __attribute__ ((used))
-    is_subdir(const char *subdir, const char *dir, size_t len)
+    is_subdir(const char *subdir, const char *dir)
 {
+	size_t dirlen = strlen(dir);
 	size_t subdirlen = strlen(subdir);
 
-	if (subdirlen < len)
+	// @dir has to be at least as long as @subdir
+	if (subdirlen < dirlen)
 		return false;
-	if (strncmp(subdir, dir, len) != 0)
+	// @dir has to be a prefix of @subdir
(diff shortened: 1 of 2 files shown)


src/mount-support.c
```diff
@@ -392,3 +392,23 @@ static char * __attribute__ ((used))
 	*offsetp = offset;
 	return (offset < fulllen) ? &path[offset] : NULL;
 }
+
+/*
+ * Check that @subdir is a subdir of @dir.
+ * @len is the length of @dir (to avoid having to recalculate it).
+**/
+static bool __attribute__ ((used))
+    is_subdir(const char *subdir, const char *dir, size_t len)
```
kyrofa: How to use this function is not obvious from its signature. You documented this (thank you), but perhaps it would be more clear to accept a length for both of these?

kyrofa: Also, this doesn't take into account `..` etc. I assume that's okay?

zyga: This function comes straight from upstream LXD codebase. I agree on the size being awkward to pass. The original code doesn't really take any advantage of this. I will change the method to compute both sizes by itself.

kyrofa: > I will change the method to compute both sizes by itself.

Yeah that's the ideal solution.

zyga: Ha, I guess this was an useful exercise. That function is buggy

src/mount-support.c
```diff
@@ -393,22 +393,34 @@ static char * __attribute__ ((used))
...
 		return false;
-	if (strncmp(subdir, dir, len) != 0)
+	// @dir has to be a prefix of @subdir
+	if (strncmp(subdir, dir, dirlen) != 0)
 		return false;
-	if (dir[len - 1] == '/')
+	// @dir can look like "path/" (that is, end with the directory separator).
+	// When that is the case then given the test above we can be sure @subdir
+	// is a real subdirectory.
+	if (dirlen > 0 && dir[dirlen - 1] == '/')
```
kyrofa, line 414: Ah yeah, good catch.

kyrofa:
:+1: looks good to me.

PR #66, opened 2016-06-30
Packaging updates

mwhudson:
Here are some changes that were made to the packaging before the package was uploaded to debian. What to put in debian/changelog was mostly just a guess, not really sure how to represent the mixed up history.

I'm also not sure about what --libexecdir should be -- /usr/lib/snapd or /usr/lib/snap-confine? I've left this unchanged for now.


zyga:
Thank you!

I'll patch the changeling, we use it for mostly informative purposes since it is not really used directly in ubuntu or debian. The vendor logic will allow me to run some of the per-commit tests on debian which is very valuable.

PR #67, opened 2016-07-01
Control apparmor and seccomp separately

zyga:
This patch allows seccomp and apparmor support to be disabled
separately. In practice we can now start enabling seccomp support
where userspace libraries are recent enough.

This doens't bring us out of devmode yet but makes a step towards
that goal. On distributions that both apparmor and seccomp are
not fully available yet the old --disable-confinement is now
spelled as --disable-seccomp --disable-apparmor.


mvo5:
:+1:

PR #69, opened 2016-07-01
Fix check for CONFINEMENT_TESTS

drizzt:
Without this fix CONFINEMENT_TESTS is always set if you build on i686


kyrofa:
Thanks @drizzt, looks like a logic error to me :+1: .

zyga:
Nice catch, thank you :)

PR #70, opened 2016-07-04
fix path in snap-confine.lintian-overrides

mwhudson:


zyga:
Nice catch, thank you!

PR #72, opened 2016-07-06
use execle() with empty environment when calling snappy-app-dev

jdstrand:
use execle() with clean environment instead of execl() when calling
snappy-add-dev to avoid influencing snappy-add-dev and allowing privilege
escalation. snappy-app-dev is only called when devices are assigned via udev
tagging, but currently snapd does not create these udev tags during 'snap
install' or 'snap connect' so this code branch is not entered during app
invocation.

Currently this is not a security issue since no snaps will be udev tagged via
snapd (as such, today this can only be exploited by manually creating the
necessary udev tags (and there is no reason or documentation to do so), but
this requires root and is thus not a privilege escalation).

This is a vulnerability in waiting though since whenever the udev tagging is
used and generated by snapd, then an unprivileged user would be able to create
a crafted environment and pass to snappy-app-dev via execl(). While
snappy-app-dev is constrained by the launcher AppArmor profile, clearly this
issue must be addressed.

Thanks to Sebastian Krahmer from SUSE for discovering this issue.

### DIFF
--- a/spread-tests/cgroup-used/task.yaml
+++ b/spread-tests/cgroup-used/task.yaml
@@ -11,7 +11,6 @@ restore: |
     # no way to clear cgroup for hello-world atm
 execute: |
     cd /
-    udevadm settle
     echo Install hello-world
     snap install hello-world
     echo Clear udev tags and cgroups with non-test device and running hello-world
@@ -19,6 +18,7 @@ execute: |
     udevadm control --reload-rules
     udevadm settle
     udevadm trigger
+    udevadm settle
     hello-world.echo | grep Hello
     echo Verify no tags for hello-world.echo for kmsg
     if udevadm info /sys/devices/virtual/mem/kmsg | grep snap_hello-world_echo ; then exit 1; fi
--- a/spread-tests/cgroup-used/task.yaml
+++ b/spread-tests/cgroup-used/task.yaml
@@ -8,7 +8,7 @@ restore: |
     udevadm settle
     udevadm trigger
     udevadm settle
-    # no way to clear cgroups atm
+    # no way to clear cgroup for hello-world atm
 execute: |
     cd /
     udevadm settle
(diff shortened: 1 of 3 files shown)


tyhicks:
Looks good to me but I'd like to see a comment before we call execle() on why the env needs to be empty (or at least not controlled by the user).

zyga:
I'd like to see a test that shows this stays fixed. Can we please work on a spread test?

zyga:
I'll create a bug report on snap-confine to link it here

zyga:
This is now bug https://bugs.launchpad.net/snap-confine/+bug/1599609 (private)

spread-tests/cgroup-used/task.yaml
```diff
@@ -0,0 +1,36 @@
...
+    udevadm control --reload-rules
+    udevadm settle
+    udevadm trigger
+    hello-world.echo | grep Hello
+    echo Verify no tags for hello-world.echo for kmsg
+    if udevadm info /sys/devices/virtual/mem/kmsg | grep snap_hello-world_echo ; then exit 1; fi
+    echo Manually add udev tags for hello-world.echo for kmsg
+    echo 'KERNEL=="kmsg", TAG+="snap_hello-world_echo"' > /etc/udev/rules.d/70-spread-test.rules
+    echo Simulate snapd udev triggers
+    udevadm control --reload-rules
```
zyga, line 28: Not 100% sure but udev tracks this nowadays so I think those reload bits are not strictly mandatory. Let's leave them but just saying.

zyga:
Thank you for the fix Jamie. As discussed on IRC I'll merge and release this. We can add more spread tests as we go but this should be sufficient to deal with the issue at hand. Thanks for the cgroup spread test, it is very useful to have real end-user tests that don't mock anything really.

spread-tests/cgroup-used/task.yaml
jdstrand, line 28: Well, I had only the udevadm commands without settle and it didn't work-- I didn't think udevadm settle was strictly required these days, but it seems to be which suggests that all the udevadm commands are needed, but, yes, we can investigate more in the future, but this doesn't hurt anything to have in the testsuite.

PR #74, opened 2016-07-07
Debian spread tests

zyga:
This patch just removes various blacklists that kept tests from running
on Jessie. Now that we have a Sid image we can run tests there.

### DIFF
--- a/spread-tests/hello-world-is-confined/task.yaml
+++ b/spread-tests/hello-world-is-confined/task.yaml
@@ -1,0 +1,13 @@
+summary: Check that hello-world.evil is confined
+# This is blacklisted on distributions without full confinement
+systems: [-debian-8, -debian-sid-grub]
+restore: |
+    snap remove hello-world
+execute: |
+    cd /
+    echo Run some hello-world stuff
+    snap install hello-world
+    hello-world.echo | grep Hello
+    hello-world.env | grep SNAP_NAME=hello-world
+    echo Ensure that we get an error if hello-world.evil does not return an error
+    if hello-world.evil; then exit 1; fi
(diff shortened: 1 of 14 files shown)


spread-tests/hello-world-is-confined/task.yaml
```diff
@@ -0,0 +1,13 @@
+summary: Check that hello-world.evil is confined
```
niemeyer, line 1: What's actually being tested here?  We're of course not very interested in hello world itself.

spread-tests/hello-world-is-confined/task.yaml
```diff
@@ -0,0 +1,13 @@
...
+# This is blacklisted on distributions without full confinement
+systems: [-debian-sid-grub]
+restore: |
+    snap remove hello-world
+execute: |
+    cd /
```
niemeyer, line 7: Seems unnecessary.

spread-tests/hello-world-is-confined/task.yaml
```diff
@@ -0,0 +1,13 @@
...
+    echo Run some hello-world stuff
+    snap install hello-world
+    hello-world.echo | grep Hello
```
niemeyer, line 10: This test may be ported to use test-snapd-tools, which offers saner apps for this sort of testing. Rather than "evil", try to exercise the system in some way that is expected to break with "cat /foo", for example.

spread-tests/hello-world-is-confined/task.yaml
niemeyer, line 1: Reading the logic, this might be "Check that confinement works", and the test name might be "confined-blocked" or similar.

zyga:
I'll close this for now. I will reopen it with new way to support Debian.

PR #75, opened 2016-07-07
add a regression test for LP: #1599608

jdstrand:
The setup for this test is unorthodox because by the time the cgroup code is executed, the mounts are in place and /lib/udev/snappy-app-dev from the OS snap is used. Unfortunately, simple bind mounts over /snap/ubuntu-core/current/lib/udev don't work and the OS snap must be unpacked, lib/udev/snappy-app-dev modified to be tested, repacked and mounted. We unmount the OS snap and move it aside to avoid both the original and the updated OS snap from being mounted on the same mount point, which confuses the kernel.

### DIFF
--- a/spread-tests/regression/lp-1599608/task.yaml
+++ b/spread-tests/regression/lp-1599608/task.yaml
@@ -37,9 +37,9 @@ execute: |
     PATH=/foo:$PATH TESTVAR=bar hello-world.env | grep PATH
     cat /run/udev/spread-test.out
     echo "Ensure user-specified PATH is not used"
-    !grep 'PATH=/foo' /run/udev/spread-test.out
+    ! grep 'PATH=/foo' /run/udev/spread-test.out
     echo "Ensure environment is clean"
-    !grep 'TESTVAR=bar' /run/udev/spread-test.out
+    ! grep 'TESTVAR=bar' /run/udev/spread-test.out
 restore: |
     echo "Remove hello-world"
     snap remove hello-world
--- a/spread-tests/regression/lp-1599608/task.yaml
+++ b/spread-tests/regression/lp-1599608/task.yaml
@@ -9,6 +9,36 @@
 summary: Check that execle doesn't regress
 # This is blacklisted on debian because we first have to get the dpkg-vendor patches
 systems: [-debian-8]
+prepare: |
+    cd /
+    echo "Install hello-world"
+    snap install hello-world
+    # all of this ls madness can go away when we have remote environment
+    # variables
+    echo "Unmount original core snap"
+    umount $(ls -1d /snap/ubuntu-core/* | grep -v current | tail -1)
+    mv $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1) $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1).orig
+    echo "Create modified core snap for snappy-app-dev"
+    unsquashfs $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap.orig | tail -1)
+    echo 'echo PATH=$PATH > /run/udev/spread-test.out' >> ./squashfs-root/lib/udev/snappy-app-dev
+    echo 'echo TESTVAR=$TESTVAR >> /run/udev/spread-test.out' >> ./squashfs-root/lib/udev/snappy-app-dev
+    mksquashfs ./squashfs-root $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap.orig | tail -1 | sed 's/.orig//') -comp xz
+    if [ ! -e $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1) ]; then exit 1; fi
+    echo "Mount modified core snap"
+    mount $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1) $(ls -1d /snap/ubuntu-core/* | grep -v current | tail -1)
+execute: |
+    echo "Add a udev tag so affected code branch is exercised"
+    echo 'KERNEL=="uinput", TAG+="snap_hello-world_env"' > /etc/udev/rules.d/70-spread-test.rules
+    udevadm control --reload-rules
+    udevadm settle
+    udevadm trigger
+    udevadm settle
+    PATH=/foo:$PATH TESTVAR=bar hello-world.env | grep PATH
+    cat /run/udev/spread-test.out
+    echo "Ensure user-specified PATH is not used"
+    !grep 'PATH=/foo' /run/udev/spread-test.out
+    echo "Ensure environment is clean"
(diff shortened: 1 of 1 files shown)


spread-tests/regression/lp-1599608/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+summary: Check that execle doesn't regress
+# This is blacklisted on debian because we first have to get the dpkg-vendor patches
+systems: [-debian-8]
+restore: |
+    echo Remove hello-world
+    snap remove hello-world
+    echo Unmount the modified OS snap
+    # all of this ls madness can go away when we have remote environment
+    # variables
+    umount $(ls -1d /snap/ubuntu-core/* | grep -v current | tail -1)
```
zyga: I think you should be able to just `unmount /snap/ubuntu-core/current`

spread-tests/regression/lp-1599608/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+    if [ "x"$(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap.orig | tail -1) != "x" ]; then mv -f $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap.orig | tail -1) $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap.orig | tail -1 | sed 's/.orig//') ; fi
+    echo Mount the original OS snap
+    mount $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1) $(ls -1d /snap/ubuntu-core/* | grep -v current | tail -1)
```
zyga: I wonder if both mount and unmount could use system for this. It's not required since this apparently works but ...

spread-tests/regression/lp-1599608/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+    rm -rf /squashfs-root
+    rm -f /run/spread-test.out
+    rm -f /etc/udev/rules.d/70-spread-test.rules
+    udevadm control --reload-rules
+    udevadm settle
+    udevadm trigger
+    udevadm settle
+execute: |
```
zyga: Can you please spit this into prepare/execute/restore parts so that it is somewhat more readable

spread-tests/regression/lp-1599608/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+    mount | grep core || true
+    echo Add a udev tag so affected code branch is exercised
+    echo 'KERNEL=="uinput", TAG+="snap_hello-world_env"' > /etc/udev/rules.d/70-spread-test.rules
+    udevadm control --reload-rules
+    udevadm settle
+    udevadm trigger
+    udevadm settle
+    PATH=/foo:$PATH TESTVAR=bar hello-world.env | grep PATH
+    cat /run/spread-test.out
+    if grep 'PATH=/foo' /run/spread-test.out ; then exit 1; fi
```
zyga: This can be replaced with:

`!grep -v 'PATH=/foo' /run/spread-test.out`

I'd also love to see something for humans like `echo "We can now ensure that PATH wasn't set to /foo"`

spread-tests/regression/lp-1599608/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+    udevadm settle
+    udevadm trigger
+    udevadm settle
+execute: |
+    cd /
+    echo Install hello-world
+    snap install hello-world
+    # all of this ls madness can go away when we have remote environment
+    # variables
+    echo Unmount original OS snap
```
zyga: Core snap :-)

spread-tests/regression/lp-1599608/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+    echo Add a udev tag so affected code branch is exercised
+    echo 'KERNEL=="uinput", TAG+="snap_hello-world_env"' > /etc/udev/rules.d/70-spread-test.rules
+    udevadm control --reload-rules
+    udevadm settle
+    udevadm trigger
+    udevadm settle
+    PATH=/foo:$PATH TESTVAR=bar hello-world.env | grep PATH
+    cat /run/spread-test.out
+    if grep 'PATH=/foo' /run/spread-test.out ; then exit 1; fi
+    if grep 'TESTVAR=bar' /run/spread-test.out ; then exit 1; fi
```
zyga: `echo "We can also ensure that other environment variables were not preserved"`

spread-tests/regression/lp-1599608/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+    mv $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1) $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1).orig
+    echo Create modified OS snap for snappy-app-dev
+    unsquashfs $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap.orig | tail -1)
+    echo 'echo PATH=$PATH > /run/spread-test.out' >> ./squashfs-root/lib/udev/snappy-app-dev
+    echo 'echo TESTVAR=$TESTVAR >> /run/spread-test.out' >> ./squashfs-root/lib/udev/snappy-app-dev
+    mksquashfs ./squashfs-root $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap.orig | tail -1 | sed 's/.orig//') -comp xz
+    if [ ! -e $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1) ]; then exit 1; fi
+    echo Mount modified OS snap
+    mount $(ls -1 /var/lib/snapd/snaps/ubuntu-core_*.snap | tail -1) $(ls -1d /snap/ubuntu-core/* | grep -v current | tail -1)
+    mount | grep core || true
```
zyga: Why do you have the `|| true` part here?

jdstrand: This was a debugging line. Removed.

spread-tests/regression/lp-1599608/task.yaml
jdstrand: Done

spread-tests/regression/lp-1599608/task.yaml
jdstrand: Done

spread-tests/regression/lp-1599608/task.yaml
jdstrand: Done

spread-tests/regression/lp-1599608/task.yaml
jdstrand: Done

spread-tests/regression/lp-1599608/task.yaml
jdstrand: What do you mean by "could use system for this"?

zyga:
This looks much nicer, thank you!

PR #76, opened 2016-07-08
Drop the alternative to ROOTFS_IS_CORE_SNAP

zyga:
Since everyone is now shipping this code let’s simplify testing and
remove the old option entirely.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -10,16 +10,16 @@ include /usr/share/dpkg/buildflags.mk
 # security of the system.  Discuss a proper approach to this for downstreams
 # if and when they approach us
 ifeq ($(shell dpkg-vendor --query Vendor),Ubuntu)
-    VENDOR_ARGS=--enable-rootfs-is-core-snap
+    VENDOR_ARGS=--enable-nvidia-ubuntu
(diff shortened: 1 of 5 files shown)


debian/rules
```diff
@@ -10,16 +10,16 @@ include /usr/share/dpkg/buildflags.mk
 # security of the system.  Discuss a proper approach to this for downstreams
 # if and when they approach us
 ifeq ($(shell dpkg-vendor --query Vendor),Ubuntu)
-    VENDOR_ARGS=--enable-rootfs-is-core-snap
+    VENDOR_ARGS=--enable-nvidia-ubuntu
 else
-    VENDOR_ARGS=--disable-confinement
+    VENDOR_ARGS=--disable-apparmor
 endif
```
jdstrand, line 16: These changes don't seem strictly related to this PR but seem ok.

jdstrand:
I've not tested this but I like removing code. The code changes LGTM.

debian/rules
zyga, line 16: No, they are a manual cherry pick from the debian branch

PR #77, opened 2016-07-14
Fix bind mounting of host filesystem and nvidia on Ubuntu

zyga:
This patch fixes the issue in which host filesystem and nvidia driver
couldn't be bind mounted. The host filesystem is now bind mounted
whenever we use --enable-rootfs-is-core-snap (i.e. always).

The issue was caused by two separate facts:
- the apparmor profile wasn't allowing the particular bind mount to
  happen
- the host filesystem wasn't mounted on ubuntu

This setup was tested earlier but due to apparmor issue where the
profile wasn't really applied it regressed and stopped working.

Fixes: https://bugs.launchpad.net/snap-confine/+bug/1602576


mvo5:
:+1:

PR #80, opened 2016-07-15
Split seccomp handling into two phases

zyga:
This patch splits the current seccomp handling code into two phases. The
first phase runs before we start all the bind mount tricks and
pivot_root. This phase has unobstructed access to /var/lib/snapd/ from
the host filesystem.

The second phase just calls seccomp_load() on a context object made
during the first phase. It happens exactly at the same time we would
have parsed and loaded the context in the original code.

In effect this lets us stop bind mounting /var/lib/snapd from the host
into the application execution environment. This also lets us see
updated /var/lib/snapd mount points (like /var/lib/snapd/hostfs) with
new releases of the core snap, regardless of how the host distribution
packaging of snapd looks like.


mvo5:
:+1:

PR #83, opened 2016-07-15
Drop /var/lib/snapd bind mount

zyga:
This patch drops the /var/lib/snapd bind mount. In effect the
/var/lib/snapd directory will appear exactly as it is in the core snap.

In effect, at least since revision 138, the core snap now exposes the
hostfs directory under /var/lib/snapd/hostfs and this can be relied upon
by the nvidia driver mount code.

The downside is that /var/lib/snapd/ will be otherwise empty (no
state.json, etc) but since there is no interface for accessing that data
and we already have the snapd-control interface, this is deemed an
acceptable compromise.


zyga:
NOTE: Don't land this until revision 138 of the core snap is promoted to stable!

PR #87, opened 2016-07-15
Allow snap-confine to bind-mount fonts from the hostfs

zyga:
This patch allows snap-confine to do a read-only bind mount of the
/usr/share/fonts directory from the host filesystem. This is done in
anticipation of support of this feature in snapd.


mvo5:
:+1:

PR #88, opened 2016-07-19
snap-confine is now on /usr/lib/snapd

drizzt:


debian/usr.bin.snap-confine
```diff
@@ -106,7 +106,7 @@
     mount options=(rw rbind) /var/tmp/ -> /tmp/snap.rootfs_*/var/tmp/,
     mount options=(rw rbind) /run/ -> /tmp/snap.rootfs_*/run/,
     mount options=(rw rbind) /media/ -> /tmp/snap.rootfs_*/media/,
-    mount options=(rw rbind) /lib/modules/ -> /tmp/snap.rootfs_*/lib/modules/,
+    mount options=(rw rbind) {/usr,}/lib/modules/ -> /tmp/snap.rootfs_*/lib/modules/,
```
zyga, line 109: This seems unrelated

drizzt, line 109: On ArchLinux /lib is a symlink to /usr/bin and /bin, /sbin, /usr/sbin are symlinks to /usr/bin.
So to make apparmor "happy" in order to bind mount /lib/modules you need to supports /usr/lib/modules too

PR #91, opened 2016-07-25
detect OpenWrt as classic distro

teknoraver:


src/classic.c
```diff
@@ -10,5 +10,6 @@ bool is_running_on_classic_distribution()
 	    || access("/var/lib/dpkg/status", F_OK) == 0
 	    || access("/var/lib/pacman", F_OK) == 0
 	    || access("/var/lib/portage", F_OK) == 0
-	    || access("/var/lib/rpm", F_OK) == 0;
+	    || access("/var/lib/rpm", F_OK) == 0
+	    || access("/sbin/procd", F_OK) == 0;
```
Conan-Kudo, line 14: Shouldn't this be detecting the opkg database path rather than `/sbin/procd`? It doesn't seem to make sense to detect an init system for detecting openwrt as a classic distro.

teknoraver, line 14: OpenWrt has a build time configuration option: "Remove ipkg/opkg status data files in final images" which can lead again to fake classic detection

Conan-Kudo, line 14: Wouldn't `/var/lib/opkg` or `/var/lib/ipkg` still exist anyway?

teknoraver, line 14: `target/imagebuilder/files/Makefile:  $(if $(CONFIG_CLEAN_IPKG),rm -rf $(TARGET_DIR)/usr/lib/opkg)`

Conan-Kudo, line 14: Would it still be a "classic" system if the database is gone?

teknoraver, line 14: the database is not mandatory, you can build all the daemons in the firmware instead of installing them at runtime

zyga:
I will take this patch for now. I will later replace all of those with a check for a file in `/var/lib/snapd/` somewhere.

EDIT: My point is that I want one codebase make the decision, not two that have to be kept in sync

PR #93, opened 2016-07-27
Bind mount /var/log for log-observe

zyga:
This patch bind mounts the /var/log directory so that snaps can use the
log-observe interface to get access to system log files.

The fix also involves another permission in the apparmor profile for the
particular mount to go ahead.

The whole set is tested with a new spread regression test.

Fixes: https://bugs.launchpad.net/snap-confine/+bug/1606277

### DIFF
--- a/spread-tests/regression/lp-1606277/task.yaml
+++ b/spread-tests/regression/lp-1606277/task.yaml
@@ -1,0 +1,17 @@
+summary: Regression check for https://bugs.launchpad.net/snap-confine/+bug/1606277 
+# This is blacklisted on debian because debian doesn't use apparmor yet
+systems: [-debian-8]
+details: |
+    A missing bind mount for /var/log prevents access to system log files
+    even if the log-observe interface is being used.
+prepare: |
+    echo "Having installed the snapd-hacker-toolbelt snap"
+    snap install snapd-hacker-toolbelt
+    echo "And having connected the log-observe interface"
+    snap connect snapd-hacker-toolbelt:log-observe ubuntu-core:log-observe
+execute: |
+    echo "We can now see a non-empty /var/log directory"
+    cd /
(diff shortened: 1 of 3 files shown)


mvo5:
:+1: please merge once tests are green. btw, thanks for the regression test!

spread-tests/regression/lp-1606277/task.yaml
```diff
@@ -0,0 +1,17 @@
...
+    even if the log-observe interface is being used.
+prepare: |
+    echo "Having installed the snapd-hacker-toolbelt snap"
+    snap install snapd-hacker-toolbelt
+    echo "And having connected the log-observe interface"
+    snap connect snapd-hacker-toolbelt:log-observe ubuntu-core:log-observe
+execute: |
+    echo "We can now see a non-empty /var/log directory"
+    cd /
+    [ "$(/snap/bin/snapd-hacker-toolbelt.busybox ls /var/log | wc -l)" != 0 ]
```
jdstrand, line 15: Is non-empty the correct check? When I was triaging this issue I saw that there were things in /var/log due to the core snap having them there. Perhaps a better test would be for /var/log/kern.log or /var/log/syslog? Those might be Debian/Ubuntu specific though. You could look in /snap/ubuntu-core/current/var/log and compare the number of files there to the number in /var/log in the snap....

zyga, line 15: Without this bind mount `/var/log` would be always empty as it would be the read-only core snap. I think we can also put a canary file in the real `/var/log` but the test as-is seems correct to me.

jdstrand, line 15: The read only core snap has files in /var/log is what I'm saying.

zyga, line 15: Oh, I didn't know that. That feels somewhat bogus though :-)

In either case I see the need for a canary test now

PR #96, opened 2016-08-03
unconditionally mount /usr/src (LP: #1597842)

jdstrand:
Some snaps need access to the kernel sources. On an Ubuntu classic system these
are found in /usr/src. This does not cover other locations for cross-distro
or kernel snaps.

To unblock people, this simply mounts /usr/src with the other OS mounts, but
this should really be interface-dependent along with /var/log. Since that
requires a bit of reorganizing, https://bugs.launchpad.net/snappy/+bug/1609499
covers this work.


zyga:
Looks good though I worry that downstream distributions that lack `/usr/src/` will all have to patch this away. Still, I'll probably merge this soon.

Conan-Kudo:
@zyga At least in Fedora, we do use `/usr/src` (`/usr/src/kernels`) for holding kernel sources for building kmods. These are available via the `kernel-devel` package.

zyga:
Merging, thank you!

PR #97, opened 2016-08-11
Restore creation of $SNAP_USER_DATA

zyga:
This patch restores creation of SNAP_USER_DATA directory. This was a regression of functionality that occurred because the new code path in snapd that does the same thing is not active yet (it depends on snap-exec).

The fix comes with spread tests that should reliably work even when snap-exec is in use.

Fixes: https://bugs.launchpad.net/snap-confine/+bug/1612120

### DIFF
--- a/src/user-support.c
+++ b/src/user-support.c
@@ -1,0 +1,110 @@
+/*
+ * Copyright (C) 2015 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#include "config.h"
+#include "user-support.h"
+
+#include <errno.h>
+#include <fcntl.h>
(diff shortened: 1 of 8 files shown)


src/user-support.c
```diff
@@ -0,0 +1,110 @@
...
+	free(path_copy);
+}
+
+void setup_user_data()
+{
+	const char *user_data = getenv("SNAP_USER_DATA");
+
+	if (user_data == NULL)
+		return;
+	// Only support absolute paths.
```
mvo5, line 104: I was initially wondering if we need extra validation here, but I think we don't because privs have already dropped when mkdir is called.

mvo5, line 104: We might check for ".." in the path though and disallow that.

mvo5:
Looks good, thanks for the tests!

src/user-support.c
zyga, line 104: We might though snap-confine is already confined with apparmor so (at least where apparmor is used) this is not critical. I will land this and follow up with a small patch to check for `..`. Thanks for spotting this.

mvo5:
Sorry, but this appears to cause a regression with the  apparmor rules on ecryptfs.

zyga:
I will try to come up with a one-time fix by testing locally with encrypted home directory but we should investigate how to support that in spread.

mvo5:
:+1: looks good, will test in a bit

jdstrand:
I am not a fan of this change because I think it is papering over the larger problem of using sudo with snap commands and the fact that it is creating directories with root:root permissions in the user's home directory. This feels very wrong and causes lots of pain for users when they use the command under sudo and it later fails when not running under sudo.

That said, the architects have not taken a stance on this issue yet and the current implementation dictates that it should work, so these rules are required.

+0 (as in, I won't block this commit, but I do not approve of the changes)

PR #99, opened 2016-08-12
Disable owner checks for 14.04-style private home

zyga:
This patch is a follow-up on the earlier work that allowed snap-confine
to access encrypted home directory even if invoked as another user (e.g.
sudo). Encrypted home directory support has evolved over time and the
patch applied earlier only covered the most recent implementation of the
concept. This patch discards owner checks for older implementations.


jdstrand:
Like the other change to remove owner, this is also fine.

+1

PR #100, opened 2016-08-12
relocate to /tmp if curdir is no longer accessible

mvo5:


zyga:
+1 on the code, I'd like to see a spread test for this and I'd like security to ack it too.

mvo5:
There was talk in the channel that we should have a chmod 0 /var/lib/snapd/void to chdir into

zyga:
Closing this since `the-void` branch has already landed.

PR #101, opened 2016-08-12
add workaround rule for LP: #1584456

jdstrand:


tyhicks:
Looks correct to me.

jdstrand:
"2016/08/12 16:32:31 Cannot allocate linode:ubuntu-16.04-64-grub: authentication failed" is the testsuite failure which is unrelated to the PR.

zyga:
+1, thank you for fixing this

fallen:
Thanks!

PR #102, opened 2016-08-15
Take upstream version from VERSION file

zyga:
This patch makes it easier to identify upstream version as well as to
increment it with scripts. The version file is also shipped in upstream
tarballs so that everything can be rebuilt correctly.


morphis:
LGTM

PR #103, opened 2016-08-15
Use downstream packaging in spread tests

zyga:
This patch changes spread setup phase to build the native package (for
Ubuntu and soon for Debian) in a way that is more consistent with the
actual release process. This should help us release with confidence as
all CI tests will run with real downstream packaging applied to real
upstream tarball and build in an isolated manner with sbuild.

This also paves the way towards the removal of the debian/ packaging
from the upstream repository. This will be done in a separate pull
request to avoid clutter.

The spread prepare stage was moved to .spread-prepare.sh so that it can
be shell-checked more easily.

The debian-8 spread target was removed because currently the downstream
packaging is not compatible. Debian maintainers were notified to the
issue so that we should be able to re-enable Debian (sid) support soon.


.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
...
+# Record where the top level directory is
+top_dir=$(pwd)
+
+# Record the current distribution release data to know what to do
+release_ID="$( . /etc/os-release && echo "${ID:-linux}" )"
+release_VERSION_ID="$( . /etc/os-release && echo "${VERSION_ID:-}" )"
+
+
+# Create source distribution tarball and place it in the top-level directory.
+create_dist_tarball() {
```
mvo5: This appears to be useful outside of spread as well, maybe it could be a toplevel `release.sh` script?

.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
+#!/bin/sh
+# This script is started by spread to prepare the execution environment
+set -x
```
mvo5: (nitpick^2) why multiple lines, `set -eux` seems to be ok?

.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
...
+
+# Record the current distribution release data to know what to do
+release_ID="$( . /etc/os-release && echo "${ID:-linux}" )"
+release_VERSION_ID="$( . /etc/os-release && echo "${VERSION_ID:-}" )"
+
+
+# Create source distribution tarball and place it in the top-level directory.
+create_dist_tarball() {
+    # Load the version number from a dedicated file
+    local pkg_version=
```
mvo5: I wonder if we could combine this and the following line to `local pkg_version=$(cat "$top_dir/VERSION)`. Using local in this way should be fine with dash https://wiki.ubuntu.com/DashAsBinSh#local - interessting enough `local` is not posix compliant, but I think we can ignore this :)

.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
...
+
+    # Ensure that build system is up-to-date and ready
+    autoreconf -i
+    # XXX: This fixes somewhat odd error when configure below (in an empty directory) fails with:
+    # configure: error: source directory already configured; run "make distclean" there first
+    test -f Makefile && make distclean
+
+    # Create a scratch space to run configure
+    scratch_dir="$(mktemp -d)"
+    trap 'rm -rf "$scratch_dir"' EXIT
```
mvo5: I see multiple `trap` usages in this script, they will override each other, a helper like https://github.com/Debian/apt/blob/master/test/integration/framework#L247 may help here for multiple traps.

.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
...
+
+    # Configure the project in a scratch directory
+    cd "$scratch_dir"
+    "$top_dir/configure" --prefix=/usr
+
+    # Create the distribution tarball
+    make dist
+
+    # Ensure we got the tarball we were expecting to see
+    test -f "snap-confine-$pkg_version.tar.gz"
```
mvo5: I think this one is redundant (nice but redundant) because the mv below will already fail if the file is not there.

.spread-prepare.sh
zyga: Sure!

.spread-prepare.sh
zyga: Just my habit, I'll combine those into one line :)

.spread-prepare.sh
zyga: I did this because of spellcheck, otherwise the error code from cat is lost. I don't mind it either way but the current code is spellcheck clean which feels good.

.spread-prepare.sh
zyga: Hmm, I assumed that traps in functions are okay (since they run in sub shells). Perhaps I was wrong. Let me check this in more detail. If anything there could be one trap for all things that need cleanup.

.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
...
+
+    # Move it to the top-level directory
+    mv "snap-confine-$pkg_version.tar.gz" "$top_dir/"
+}
+
+
+# build Ubuntu binary package and place it in $top_dir
+# $1 = /etc/os-release ID field
+# $2 = /etc/os-release VERSION_ID field (possibly empty!)
+build_debian_or_ubuntu_package() {
```
mvo5: This function is very long. This makes me unhappy, I'm exploring some ideas right now what could be done about it.

.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
...
+            esac
+            ;;
+        *)
+            echo "unsupported distribution ID: $1"
+            exit 1
+            ;;
+    esac
+
+    # Ensure that we have a sbuild chroot ready
+    if ! schroot -l | grep "chroot:${distro_codename}-.*-sbuild"; then
```
mvo5: While sbuild is as close as you can get to the real buildds, I'm not sure if its not overkill. given that we run in a fresh linode machine and all. That would also allow us to remove the sbuild keys. But maybe going too far here, happy to discuss this.

.spread-prepare.sh
zyga: Yep, the move was added later (lots of internal cycles on this script). I'l get rid of it.

.spread-prepare.sh
zyga: This just keeps us honest. I think if the price (in terms of time spent) is small enough it is just the right thing to do. Fedora build will similarly use mock (which takes around a minute) to build the package.

The goal is to avoid "uhhh" moments and stress when we actually release.

.spread-prepare.sh
```diff
@@ -0,0 +1,226 @@
...
+
+    # Create a scratch space 
+    scratch_dir="$(mktemp -d)"
+    trap 'rm -rf "$scratch_dir"' EXIT
+
+    # Do everything in the scratch directory
+    cd "$scratch_dir"
+
+    # Fetch the current Ubuntu packaging for the appropriate release
+    git clone -b "$distro_packaging_git_branch" "$distro_packaging_git" distro-packaging
```
mvo5: We can probably use `gbp buildpackage` here (which is hopefully closer to what we will do on a real release).

.spread-prepare.sh
mvo5: Oh, indeed, thanks. That is a good reason.

mvo5:
Thnaks for merging my changes! I think it would be nice to extract `release.sh` (or similar name, whatever you prefer) and then we can continue to iterate. I think its getting nicer already.

PR #104, opened 2016-08-16
Support recent prctl constants on old libc

vosst:
Make sure that snap-confine works when built on older libc versions lacking PR_ constants from prctl.h.


zyga:
Nice! I will fix spread API key and ask you to merge with master once and we'll land this.

zyga:
+1, merging

PR #105, opened 2016-08-16
source distro specific data

mvo5:
WDYT?


zyga:
Yes, totally! I'll work on ensuring there's some error handing for unknown distributions and releases.

Thank you for re-factoring this Mvo :-)

PR #106, opened 2016-08-16
make spread happy again

mvo5:
Moving things down one level in spread or it gets unhappy.


mvo5:
Hm, tests fail with "authentication error" :/

mvo5:
Fwiw, this works with my local qemu spread backend (but there appears to be some race, I also got failures for identical runs)

zyga:
@mvo5 spread auth error is because of how top-level spread.yaml is set up. I'll try to correct this today.

PR #107, opened 2016-08-16
Move to /var/lib/snapd/void instead of aborting

zyga:
This patch changes snap-confine behavior when the current working
directory cannot be preserved. In the past, when it was not possible to
restore the working directory after chrooting into the core snap,
snap-confine would fail with an error message.

This behavior was hard to understand so instead snap-confine will now
chdir to /var/lib/snapd/void. This directory is "chmod 0" so that for
commands that care about the current working directory, they will see
permission denied errors. This is more naturally in line with what users
would expect.


mvo5:
:+1:

PR #108, opened 2016-08-16
Remove packaging

zyga:
This branch removes the debian packaging from the tree.

This copy of packaging was not really used by Debian or by Ubuntu. We've recently switched to using downstream packaging for CI tests so keeping old packaging around is useless.

One of the files, the one that describes syscalls, is retained but moved elsewhere.


mvo5:
:+1:

PR #113, opened 2016-08-22
Allow sharing of entire snaps with the content interface

zyga:
This patch tweaks the apparmor profile of snap-confine to allow the
content interface to share an entire snap (mount it) into a directory of
another snap. In the past the apparmor profile was crafted in a way that
only subdirectory could be shared.

The patch contains a spread test that ensures this is working as
expected.

Fixes: https://bugs.launchpad.net/snap-confine/+bug/1615113


mvo5:
:+1:

PR #116, opened 2016-08-22
Add experimental spread setup for trusty.

vosst:


spread.yaml
```diff
@@ -22,6 +22,18 @@ prepare: |
     echo "Spread is running as $(id)"
     [ "$REUSE_PROJECT" != 1 ] || exit 0
     ./spread-tests/spread-prepare.sh
+    # trusty support is under development right now
+    # we special-case the release until we have officially landed
+    if ["$(lsb_release -sc)" == "trusty"]
```
zyga: `lsb_release` is not preinstalled everywhere, please move this to `spread-prepare.sh` where we already have all the information `release_ID` and `release_VERSION_ID`. Please do that before we `apt-get install snapd`.

spread.yaml
```diff
@@ -22,6 +22,18 @@ prepare: |
...
+    then
+        add-apt-repository ppa:thomas-voss/trusty
+        apt update && apt upgrade
+        apt install systemd snapd
```
zyga: systemd should be a dependency of snapd but I'm okay if that's just a temporary thing.

spread.yaml
```diff
@@ -22,6 +22,18 @@ prepare: |
...
+        # he following two lines are temporary and meant to go away
+        # once we have finalized trusty support.
+        service systemd start
+        apparmor_parser -R /etc/apparmor.d/usr.lib.snapd.snap-confine
```
zyga: Why do you need this here? Trusty support should be good in the tree now .

vosst: c'n'p leftover, good catch.

spread.yaml
vosst: Yup, working on getting the dependency in place correctly.

PR #120, opened 2016-08-26
Fix regression test when re-exec kicks in

zyga:
This patch adjusts a regression test that relied on unmounting and
changing the core snap so that before the unmount operation is performed
snapd.service and snapd.refresh.timer are both stopped. Appropriate
actions are taken to restore them later. This allows the unmount
operation to complete.


chipaca:
:+1:

PR #121, opened 2016-08-26
Allow the use of capabilities over setuid bit

zyga:
This patch adds a build-time configuration option
"--enable-caps-over-setuid" that allows distributors that wish to use
filesystem capabilities to do that instead of using setuid root
executables.

The actual change disables a fragment that checked if the user runs as
root and tweaks installation. It got minimal testing on Fedora 24.

Fixes: https://bugs.launchpad.net/snap-confine/+bug/1615610

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -148,5 +148,18 @@ AC_ARG_WITH([snap-mount-dir],
 AC_SUBST(SNAP_MOUNT_DIR)
 AC_DEFINE_UNQUOTED([SNAP_MOUNT_DIR], "${SNAP_MOUNT_DIR}", [Location of the snap mount points])
 
+AC_ARG_ENABLE([caps-over-setuid],
+    AS_HELP_STRING([--enable-caps-over-setuid], [Use capabilities rather than setuid bit]),
+    [case "${enableval}" in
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
:+1:

jdstrand:
This should be fine. I wanted to make sure that AT_SECURE was still set when using 'setcap cap_sys_admin=pe ...' since with recent changes to use 'change_profile unsafe...' we are relying secure exec to be set so the the variables are be cleared. Simple test:

```
$ cat <<EOF > print_at_secure.c
#include <stdio.h>
#include <sys/auxv.h>

int main(void)
{
 printf("AT_SECURE = %lu\n", getauxval(AT_SECURE));
 return 0;
}
EOF
$ gcc -o print_at_secure print_at_secure.c
$ sudo chown root:root ./print_at_secure
$ ./print_at_secure 
AT_SECURE = 0
$ sudo chmod 4755 ./print_at_secure
$ ./print_at_secure 
AT_SECURE = 1
$ sudo chmod 0755 ./print_at_secure
$ sudo setcap cap_sys_admin=pe ./print_at_secure
$ ./print_at_secure 
AT_SECURE = 1
```

PR #122, opened 2016-08-31
Replace pivot_root implementation by LXC's

stgraber:
This resolves the main issue we're running into with LXD when using snap-confine.

Please note that as indicated in the code and commit, this is a copy/paste from the LXC implementation of pivot_root. I am not the author of this code and LXC copyright is owned by the individual contributors (no CLA).

I'm not sure if snap-confine is released under the Canonical CLA. If it is, then this pull request is unsuitable for inclusion and you'll unfortunately have to come up with a clean room implementation of this in order to respect the copyright of the LXC contributors.

Without going too deep in the LXC git history, the copyright holders for that function appear to be:
- Daniel Lezcano (IBM at the time)
- Serge Hallyn (Canonical at the time)
- Andy Lutomirski
- Michael Holtz

As I'm not one of them, I am NOT signing off on this commit.

### DIFF
--- a/src/mount-support.c
+++ b/src/mount-support.c
@@ -26,6 +26,7 @@
 #include <sys/types.h>
 #include <sys/syscall.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <sched.h>
 #include <string.h>
 #include <mntent.h>
@@ -276,14 +277,47 @@ void setup_snappy_os_mounts()
 	// moved all paths to /tmp/snap.rootfs_*. Because we are using unshare with
 	// CLONE_NEWNS we can essentially use pivot_root just like chroot but this
 	// makes apparmor unaware of the old root so everything works okay.
+
+	// The code below was copied and adapted from src/lxc/conf.c in
+	// github.com/lxc/lxc. It is available under the terms of the
+	// LGPLv2.1 and higher and was written by various LXC developers.
+
 	debug("chrooting into %s", rootfs_dir);
-	if (chdir(rootfs_dir) == -1) {
-		die("cannot change working directory to %s", rootfs_dir);
+
+	int oldroot = -1, newroot = -1;
+	oldroot = open("/", O_DIRECTORY | O_PATH);
+	if (oldroot < 0) {
+		die("cannot open the root directory of the old filesystem");
 	}
-	if (syscall(SYS_pivot_root, ".", rootfs_dir) == -1) {
+
+	newroot = open(rootfs_dir, O_DIRECTORY | O_PATH);
+	if (newroot < 0) {
+		die("cannot open the root directory of the new filesystem");
+	}
+
+	if (fchdir(newroot)) {
+		die("cannot switch to the root directory of the new filesystem");
+	}
+
+	if (syscall(SYS_pivot_root, ".", ".") == -1) {
 		die("cannot pivot_root to the new root filesystem");
 	}
-	// Reset path as we cannot rely on the path from the host OS to
+
+	if (fchdir(oldroot) < 0) {
+		die("cannot switch to the root directory of the old filesystem");
+	}
+
+	if (umount2(".", MNT_DETACH) < 0) {
+		die("cannot unmount the old filesystem");
+	}
+
+	if (fchdir(newroot) < 0) {
+		die("cannot switch back to the root directory of the new filesystem");
+	}
+
+	close(oldroot);
+	close(newroot);
+
 	// make sense. The classic distribution may use any PATH that makes
 	// sense but we cannot assume it makes sense for the core snap
 	// layout. Note that the /usr/local directories are explicitly


stgraber:
The travis failure appears to be unrelated to this code change (authentication failure).

src/mount-support.c
```diff
@@ -276,14 +277,47 @@ void setup_snappy_os_mounts()
...
-	if (chdir(rootfs_dir) == -1) {
-		die("cannot change working directory to %s", rootfs_dir);
+
+	int oldroot = -1, newroot = -1;
+	oldroot = open("/", O_DIRECTORY | O_RDONLY);
+	if (oldroot < 0) {
+		die("failed to open the old root filesystem");
+	}
+
+	newroot = open(rootfs_dir, O_DIRECTORY | O_RDONLY);
```
zyga: O_PATH perhaps, it allows exactly what we need to do

src/mount-support.c
```diff
@@ -276,14 +277,47 @@ void setup_snappy_os_mounts()
...
+	// The code below was taken fomr src/lxc/conf.c in github.com/lxc/lxc
+	// It is available under the terms of the LGPLv2.1 and higher and
+	// was written by various LXC developers.
+
 	debug("chrooting into %s", rootfs_dir);
-	if (chdir(rootfs_dir) == -1) {
-		die("cannot change working directory to %s", rootfs_dir);
+
+	int oldroot = -1, newroot = -1;
+	oldroot = open("/", O_DIRECTORY | O_RDONLY);
```
zyga: O_PATH as well

src/mount-support.c
```diff
@@ -276,14 +277,47 @@ void setup_snappy_os_mounts()
...
+	if (oldroot < 0) {
+		die("failed to open the old root filesystem");
```
zyga: `"cannot open the root directory of the old filesystem"`

src/mount-support.c
```diff
@@ -276,14 +277,47 @@ void setup_snappy_os_mounts()
...
+	}
+
+	newroot = open(rootfs_dir, O_DIRECTORY | O_RDONLY);
+	if (newroot < 0) {
+		die("failed to open the new root filesystem");
```
zyga: `"cannot open the root directory of the new filesystem"`

src/mount-support.c
```diff
@@ -276,14 +277,47 @@ void setup_snappy_os_mounts()
...
+	}
+
+	if (fchdir(newroot)) {
+		die("failed to switch to the new root filesystem");
```
zyga: `"cannot switch to the root directory of the new filesystem"`

src/mount-support.c
```diff
@@ -276,14 +277,47 @@ void setup_snappy_os_mounts()
...
+
+	// The code below was taken fomr src/lxc/conf.c in github.com/lxc/lxc
+	// It is available under the terms of the LGPLv2.1 and higher and
+	// was written by various LXC developers.
+
 	debug("chrooting into %s", rootfs_dir);
-	if (chdir(rootfs_dir) == -1) {
-		die("cannot change working directory to %s", rootfs_dir);
+
+	int oldroot = -1, newroot = -1;
```
zyga, line 287: I like to use `__attribute__((cleanup(sc_cleanup_fd)))` though I don't think it is strictly needed in this case.

src/mount-support.c
stgraber: The problem with O_PATH is that it didn't exist at all until 2.6.39 and didn't support close() or fchdir() which I use below until the 3.5 kernel.

If snap-confine isn't expected to run with a kernel below 3.5 (say random Android kernels which were based on 3.1 or 3.4), then O_PATH would indeed be fine.

src/mount-support.c
stgraber: And same comment as to why we didn't use it in the LXC implementation (requires a 3.5+ kernel).

src/mount-support.c
zyga: I see, thanks. I think the current bare-minimum version is 3.10. Until this restriction is lifted I think it is fine to rely on 3.10-specific features.

src/mount-support.c
stgraber: Yeah and I think we very much want to be in control as to when we close those fds at the end as that's when the kernel will actually flush the old mount tree.

As for the error paths, well, we're about to exit, so closing the fds is a bit of a waste of time.

src/mount-support.c
stgraber: Ok, changed to O_DIRECTORY | O_PATH then

Conan-Kudo:
@stgraber This project, like all projects under the "snapcore" organization, are subject to Canonical's CLA.

zyga:
I'm doing large changes to this area in connection with namespace sharing feature. I'll make sure we detach the old root filesystem. I'll close this branch for now.

tianon:
@zyga just for clarity, is #126 the PR with the changes you were referring to which should also fix this too?

zyga:
@tianon this is the large pull request but the change has landed separately in #127

PR #123, opened 2016-09-02
Add mountinfo module

zyga:
This patch adds module capable of parsing /proc/self/mountinfo.

Snap-confine needs to parse this file to understand what is currently mounted in a way
that makes bind mounts visible (classic /proc/mounts is insufficient for this purpose).
For extra details about this topic see: https://lwn.net/Articles/265920/

### DIFF
--- a/src/mountinfo.c
+++ b/src/mountinfo.c
@@ -202,8 +202,8 @@ static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
 	// it to the allocated entry structure.
 	//
 	// The parsing code below, specifically parse_next_string_field(), uses
-	// this extra memory to hold data parsed from the original line. In the end
-	// the result is similar to using strtok except that the source and
+	// this extra memory to hold data parsed from the original line. In the
+	// end, the result is similar to using strtok except that the source and
 	// destination buffers are separate.
 	struct mountinfo_entry *entry =
 	    calloc(1, sizeof *entry + strlen(line) + 1);
--- a/src/mountinfo.c
+++ b/src/mountinfo.c
@@ -192,19 +192,19 @@ struct mountinfo *parse_mountinfo(const char *fname)
 
 static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
 {
-	// NOTE: the mouninfo structure is allocated along with enough extra storage
-	// to hold the whole line we are parsing. This is used as backing store for
-	// all of the text fields.
+	// NOTE: the mountinfo structure is allocated along with enough extra
+	// storage to hold the whole line we are parsing. This is used as backing
+	// store for all of the text fields.
 	//
 	// The idea is that since the line has a given length and we are only after
-	// set of substrings we can easily predict the amount of required data
+	// set of substrings we can easily predict the amount of required space
 	// (after all, it is just a set of non-overlapping substrings) and append
 	// it to the allocated entry structure.
 	//
-	// The parsing code below, specifically parse_next_string_field() uses this
-	// extra memory to hold data parsed from the original line. In the end the
-	// result is similar to using strtok except that the source and destination
-	// buffers are separate.
+	// The parsing code below, specifically parse_next_string_field(), uses
+	// this extra memory to hold data parsed from the original line. In the end
+	// the result is similar to using strtok except that the source and
+	// destination buffers are separate.
 	struct mountinfo_entry *entry =
 	    calloc(1, sizeof *entry + strlen(line) + 1);
 	if (entry == NULL) {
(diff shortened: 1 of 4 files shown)


src/mountinfo.c
```diff
@@ -0,0 +1,276 @@
...
+		goto fail;
+	if ((entry->super_opts = parse_next_string_field()) == NULL)
+		goto fail;
+	return entry;
+ fail:
+	free(entry);
+	return NULL;
+}
+
+void cleanup_mountinfo(struct mountinfo **ptr)
```
tyhicks, line 263: You didn't use this function.

src/mountinfo.c
```diff
@@ -0,0 +1,276 @@
...
+			last->next = entry;
+		} else {
+			info->first = entry;
+		}
+		last = entry;
+	}
+	return info;
+}
+
+static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
```
tyhicks, line 193: As I mentioned in IRC, this function is much more complicated than I'd like it to be. However, it looks to be correct and I don't have a better suggestion at this time.

src/mountinfo.c
```diff
@@ -0,0 +1,276 @@
...
+	nscanned = sscanf(line, "%d %d %u:%u %n",
+			  &entry->mount_id, &entry->parent_id,
+			  &entry->dev_major, &entry->dev_minor, &offset);
+	if (nscanned != 4)
+		goto fail;
+	total_offset += offset;
+	int total_used = 0;
+	char *parse_next_string_field() {
+		char *field = &entry->line_buf[0] + total_used;
+		nscanned = sscanf(line + total_offset, "%s %n", field, &offset);
```
tyhicks, line 224: Could you add a comment about what's going on here with field? It takes a while to realize that the parsed string is being stored in the array at the end of mountinfo_entry, which is being allocated to the size of the passed in line. It is also worth mentioning that the strings stored in that array are delimited with NUL characters from the calloc(). This is quite confusing and is going to take anyone reading the code quite a bit a of time to digest without a comment.

src/mountinfo.c
```diff
@@ -0,0 +1,276 @@
...
+	char *root;
+	char *mount_dir;
+	char *mount_opts;
+	char *optional_fields;
+	char *fs_type;
+	char *mount_source;
+	char *super_opts;
+
+	struct mountinfo_entry *next;
+	char line_buf[0];	// Buffer holding all of the text data above;
```
tyhicks: How about a comment saying not to add any new fields after .line_buf? Someone not well versed in C might not understand.

tyhicks:
Ack after adding the requested comments. Thanks!

src/mountinfo.c
zyga: This is a library function (for users of mount info), note that free_mountinfo is static/private.

src/mountinfo.c
zyga: Ack, I'll document this explicitly.

src/mountinfo.c
zyga: Ack

zyga: Done

src/mountinfo.c
zyga: Done

tyhicks:
Thanks for the added comments. I'm happy with this PR.

src/mountinfo.c
```diff
@@ -192,19 +192,19 @@ struct mountinfo *parse_mountinfo(const char *fname)
...
+	// set of substrings we can easily predict the amount of required space
 	// (after all, it is just a set of non-overlapping substrings) and append
 	// it to the allocated entry structure.
 	//
-	// The parsing code below, specifically parse_next_string_field() uses this
-	// extra memory to hold data parsed from the original line. In the end the
-	// result is similar to using strtok except that the source and destination
-	// buffers are separate.
+	// The parsing code below, specifically parse_next_string_field(), uses
+	// this extra memory to hold data parsed from the original line. In the end
```
Conan-Kudo: comma missing after "In the end".

src/mountinfo.c
```diff
@@ -192,19 +192,19 @@ struct mountinfo *parse_mountinfo(const char *fname)
 
 static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
 {
-	// NOTE: the mouninfo structure is allocated along with enough extra storage
-	// to hold the whole line we are parsing. This is used as backing store for
-	// all of the text fields.
+	// NOTE: the mountinfo structure is allocated along with enough extra
+	// storage to hold the whole line we are parsing. This is used as backing
+	// store for all of the text fields.
```
Conan-Kudo: "all of the text fields" can be simplified to "all text fields".

Conan-Kudo:
Aside from the couple of comments I made about your typo fixes, it looks good to me. 👍

Conan-Kudo:
👍

PR #124, opened 2016-09-05
Remove test producing spurious results on trusty.

vosst:
Seems to works fine on i386 and armhf, fails regularly on amd64.


jdstrand:
I'd prefer to fix the tests rather than removing them. How are they failing? What are the steps to reproduce (eg, run in a trusty vm? something else?)

zyga:
I've closed this. I think we can see why this fails and if it really does fail still with the new 14.04 support patches that landed around the stack.

PR #126, opened 2016-09-06
Add support module for namespace sharing

zyga:
This patch adds a module that implements mount namespace sharing. It is
not yet used by snap-confine. The module implements a set of library
functions that can perform the necessary magic.

As the topic is a bit unexpected a brief description follows.

Linux namespaces can be created with the unshare(2) call. They are
represented as files in the nsfs filesystem under /proc/$pid/ns/. The
mount namespace is represented as /proc/$pid/ns/mnt. An open file
descriptor to a file like that can be used along with setns(2) to move a
thread to a different namespace.

While namespaces are normally bound to the lifetime of a given process
they can be preserved by bind-mounting the appropriate namespace file to
another location. This feature serves as the basis for the namespace
sharing feature.

The mount namespace is a little bit special as it requires some
additional things to be in place before the bind mount can happen. All
violations of the rules listed below results in mount failing with
EINVAL.

In order to preserve the mount namespace the calling process must be in
a different mount namespace (snap-confine just uses the original namespace in
which it executes). This can be achieved by forking a child process that
can see its parent mount namespace file in /proc/$ppid/ns/mnt and having
the parent process move to a new namespace by calling unshare(). The
destination of the bind mount must be on a filesystem that is mounted
without any peers (in the sense of shared subtrees). This can be checked
by inspecting and parsing the /proc/self/mountinfo file. The new module
includes support function that bind mounts the target directory over
itself and the converts it to a private mount.

Actual mount namespaces are kept in /run/snapd/ns (this is also the
directory that gets converted to a private mount). The actual namespaces
are put in "groups" with names identical to the snap name. In practice
the preserved namespaces are in /run/snapd/ns/$SNAP_NAME.mnt

In order to make everything race-free the library uses locking based on
flock(2). There are two lock files. One global, required to make
/run/snapd/ns a private mount, and one local, required to manipulate a
particular mount namespace. The locks are /run/snapd/ns/.lock and
/run/snaps/ns/$SNAP_NAME.lock respectively.

Everything is coded defensively, terminating the process in case
something bad happens. The child process of snap-confine is using
prctl() to ensure it gets killed when the parent process dies for any
reason.

One notable thing that is _not_ present in this patch is the adjusted
apparmor profile for snap-confine and for its child process (it switches
hats to a new profile for added paranoia and security). It will be
presented along with a, hopefully small, patch that enables namespace
sharing in practice.

### DIFF
--- a/docs/snap-confine.rst
+++ b/docs/snap-confine.rst
@@ -88,9 +88,9 @@ quirks:
 Sharing of the mount namespace
 ------------------------------
 
-As a part of application startup `snap-confine` will move the process to a new
-mount namespace. Since version 1.0.41 all the applications belonging to a given
-snap will share the mount namespace amongst them.
+As of version 1.0.41 all the applications from the same snap will share the
+same mount namespace. Applications from different snaps continue to use
+separate mount namespaces.
 
 ENVIRONMENT
 ===========
@@ -133,17 +133,40 @@ FILES
 
 	Description of the seccomp profile.
 
+`/run/snapd/ns/`:
+
+    Directory used to keep shared mount namespaces.
+
+    `snap-confine` internally converts this directory to a private bind mount.
+    Semantically the behavior is identical to the following mount commands:
+
+    mount --bind /run/snapd/ns /run/snapd/ns
+    mount --make-private /run/snapd/ns
+
 `/run/snapd/ns/.lock`:
 
-	Global lock acquired to initialize namespace sharing feature
+    A `flock(2)`-based lock file acquired to create and convert
+    `/run/snapd/ns/` to a private bind mount.
 
 `/run/snapd/ns/$SNAP_NAME.lock`:
 
-    Lock acquired to initialize the namespace of a specific snap.
+    A `flock(2)`-based lock file acquired to create or join the mount namespace
+    represented as `/run/snaps/ns/$SNAP_NAME.mnt`.
 
 `/run/snapd/ns/$SNAP_NAME.mnt`:
 
-    Location of a preserved mount namespace.
+    This file can be either:
+
+    - An empty file that may be seen before the mount namespace is preserved or
+      when the mount namespace is unmounted.
+    - A file belonging to the `nsfs` file system, representing a fully
+      populated mount namespace of a given snap. The file is bind mounted from
+      `/proc/self/ns/mnt` from the first process in any snap.
+
+`/proc/self/mountinfo`:
+
+    This file is read to decide if `/run/snapd/ns/` needs to be created and
+    converted to a private bind mount, as described above.
 
 Note that the apparmor profile is external to `snap-confine` and is loaded
 directly into the kernel. The actual apparmor profile is managed by `snapd`.
--- a/docs/snap-confine.rst
+++ b/docs/snap-confine.rst
@@ -85,6 +85,13 @@ quirks:
   the execution environment. This allows various snaps, while running in
   devmode, to access the LXD socket. LP: #1613845
 
+Sharing of the mount namespace
+------------------------------
+
+As a part of application startup `snap-confine` will move the process to a new
+mount namespace. Since version 1.0.41 all the applications belonging to a given
+snap will share the mount namespace amongst them.
+
 ENVIRONMENT
 ===========
 
@@ -125,6 +132,18 @@ FILES
 `/var/lib/snapd/seccomp/profiles/*`:
 
 	Description of the seccomp profile.
+
+`/run/snapd/ns/.lock`:
+
+	Global lock acquired to initialize namespace sharing feature
+
+`/run/snapd/ns/$SNAP_NAME.lock`:
+
+    Lock acquired to initialize the namespace of a specific snap.
+
+`/run/snapd/ns/$SNAP_NAME.mnt`:
+
+    Location of a preserved mount namespace.
 
 Note that the apparmor profile is external to `snap-confine` and is loaded
 directly into the kernel. The actual apparmor profile is managed by `snapd`.
--- a/src/mountinfo.c
+++ b/src/mountinfo.c
@@ -235,7 +235,7 @@ static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
 	if ((entry->mount_opts = parse_next_string_field()) == NULL)
 		goto fail;
 	entry->optional_fields = &entry->line_buf[0] + total_used++;
-	// NOTE: This ensure s that optional_fields is never NULL.
+	// NOTE: This ensures that optional_fields is never NULL.
 	strcpy(entry->optional_fields, "");
 	for (;;) {
 		char *opt_field = parse_next_string_field();
--- a/src/mountinfo.c
+++ b/src/mountinfo.c
@@ -235,6 +235,7 @@ static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
 	if ((entry->mount_opts = parse_next_string_field()) == NULL)
 		goto fail;
 	entry->optional_fields = &entry->line_buf[0] + total_used++;
+	// NOTE: This ensure s that optional_fields is never NULL.
 	strcpy(entry->optional_fields, "");
 	for (;;) {
 		char *opt_field = parse_next_string_field();
--- a/src/mountinfo.c
+++ b/src/mountinfo.c
@@ -235,7 +235,8 @@ static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
 	if ((entry->mount_opts = parse_next_string_field()) == NULL)
 		goto fail;
 	entry->optional_fields = &entry->line_buf[0] + total_used++;
-	// NOTE: This ensures that optional_fields is never NULL.
+	// NOTE: This ensures that optional_fields is never NULL. If this changes,
+	// must adjust all callers of parse_mountinfo_entry() accordingly.
 	strcpy(entry->optional_fields, "");
 	for (;;) {
 		char *opt_field = parse_next_string_field();
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -122,6 +122,36 @@ static void test_sc_open_ns_group()
 		      (lock_file, G_FILE_TEST_EXISTS | G_FILE_TEST_IS_REGULAR));
 }
 
+static void test_sc_lock_ns_mutex_precondition()
+{
+	sc_test_use_fake_ns_dir();
+	if (g_test_subprocess()) {
+		struct sc_ns_group *group = sc_alloc_ns_group();
+		g_test_queue_free(group);
+		// Try to lock the mutex, this should abort because we never opened the
+		// lock file and don't have a valid file descriptor.
+		sc_lock_ns_mutex(group);
+		return;
+	}
+	g_test_trap_subprocess(NULL, 0, 0);
+	g_test_trap_assert_failed();
+}
+
+static void test_sc_unlock_ns_mutex_precondition()
+{
+	sc_test_use_fake_ns_dir();
+	if (g_test_subprocess()) {
+		struct sc_ns_group *group = sc_alloc_ns_group();
+		g_test_queue_free(group);
+		// Try to unlock the mutex, this should abort because we never opened the
+		// lock file and don't have a valid file descriptor.
+		sc_unlock_ns_mutex(group);
+		return;
+	}
+	g_test_trap_subprocess(NULL, 0, 0);
+	g_test_trap_assert_failed();
+}
+
 // Check that locking a namespace actually flock's the mutex with LOCK_EX
 static void test_sc_lock_unlock_ns_mutex()
 {
@@ -226,6 +256,10 @@ static void __attribute__ ((constructor)) init()
 	g_test_add_func("/ns/sc_init_ns_group", test_sc_open_ns_group);
 	g_test_add_func("/ns/sc_lock_unlock_ns_mutex",
 			test_sc_lock_unlock_ns_mutex);
+	g_test_add_func("/ns/sc_lock_ns_mutex/precondition",
+			test_sc_lock_ns_mutex_precondition);
+	g_test_add_func("/ns/sc_unlock_ns_mutex/precondition",
+			test_sc_unlock_ns_mutex_precondition);
 	g_test_add_func("/system/ns/sc_is_ns_group_dir_private",
 			test_sc_is_ns_group_dir_private);
 	g_test_add_func("/system/ns/sc_initialize_ns_groups",
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -157,18 +157,22 @@ static void test_sc_is_ns_group_dir_private()
 	const char *ns_dir = sc_test_use_fake_ns_dir();
 	g_test_queue_destroy(unmount_dir, (char *)ns_dir);
 
-	// The temporary directory should not be private initially
-	g_assert_false(sc_is_ns_group_dir_private());
+	if (g_test_subprocess()) {
+		// The temporary directory should not be private initially
+		g_assert_false(sc_is_ns_group_dir_private());
 
-	/// do what "mount --bind /foo /foo; mount --make-private /foo" does.
-	int err;
-	err = mount(ns_dir, ns_dir, NULL, MS_BIND, NULL);
-	g_assert_cmpint(err, ==, 0);
-	err = mount(NULL, ns_dir, NULL, MS_PRIVATE, NULL);
-	g_assert_cmpint(err, ==, 0);
+		/// do what "mount --bind /foo /foo; mount --make-private /foo" does.
+		int err;
+		err = mount(ns_dir, ns_dir, NULL, MS_BIND, NULL);
+		g_assert_cmpint(err, ==, 0);
+		err = mount(NULL, ns_dir, NULL, MS_PRIVATE, NULL);
+		g_assert_cmpint(err, ==, 0);
 
-	// The temporary directory should now be private
-	g_assert_true(sc_is_ns_group_dir_private());
+		// The temporary directory should now be private
+		g_assert_true(sc_is_ns_group_dir_private());
+	}
+	g_test_trap_subprocess(NULL, 0, G_TEST_SUBPROCESS_INHERIT_STDERR);
+	g_test_trap_assert_passed();
 }
 
 static void test_sc_initialize_ns_groups()
@@ -177,24 +181,34 @@ static void test_sc_initialize_ns_groups()
 		g_test_skip("this test needs to run as root");
 		return;
 	}
+	// NOTE: this is g_test_subprocess aware!
 	const char *ns_dir = sc_test_use_fake_ns_dir();
 	g_test_queue_destroy(unmount_dir, (char *)ns_dir);
+	if (g_test_subprocess()) {
+		// Initialize namespace groups using a fake directory.
+		sc_initialize_ns_groups();
 
-	// Initialize namespace groups using a fake directory.
-	sc_initialize_ns_groups();
+		// Check that the fake directory is now a private mount.
+		g_assert_true(sc_is_ns_group_dir_private());
 
-	// Check that the lock file did not leak unclosed.
+		// Check that the lock file did not leak unclosed.
 
-	// Construct the name of the lock file
-	char *lock_file __attribute__ ((cleanup(sc_cleanup_string))) = NULL;
-	lock_file = g_strdup_printf("%s/%s", ns_dir, SC_NS_LOCK_FILE);
-	// Attempt to open and lock the lock file.
-	int lock_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
-	lock_fd = open(lock_file, O_RDWR | O_CLOEXEC | O_NOFOLLOW);
-	g_assert_cmpint(lock_fd, !=, -1);
-	// The non-blocking lock operation should not fail
-	int err = flock(lock_fd, LOCK_EX | LOCK_NB);
-	g_assert_cmpint(err, ==, 0);
+		// Construct the name of the lock file
+		char *lock_file __attribute__ ((cleanup(sc_cleanup_string))) =
+		    NULL;
+		lock_file =
+		    g_strdup_printf("%s/%s", sc_ns_dir, SC_NS_LOCK_FILE);
+		// Attempt to open and lock the lock file.
+		int lock_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
+		lock_fd = open(lock_file, O_RDWR | O_CLOEXEC | O_NOFOLLOW);
+		g_assert_cmpint(lock_fd, !=, -1);
+		// The non-blocking lock operation should not fail
+		int err = flock(lock_fd, LOCK_EX | LOCK_NB);
+		g_assert_cmpint(err, ==, 0);
+		return;
+	}
+	g_test_trap_subprocess(NULL, 0, G_TEST_SUBPROCESS_INHERIT_STDERR);
+	g_test_trap_assert_passed();
 }
 
 static void __attribute__ ((constructor)) init()
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -134,55 +134,54 @@ static void test_sc_lock_unlock_ns_mutex()
 
 static void unmount_dir(void *dir)
 {
-    umount(dir);
+	umount(dir);
 }
 
 static void test_sc_is_ns_group_dir_private()
 {
-    if (geteuid() != 0) {
-        g_test_skip("this test needs to run as root");
-        return;
-    }
+	if (geteuid() != 0) {
+		g_test_skip("this test needs to run as root");
+		return;
+	}
 	const char *ns_dir = sc_test_use_fake_ns_dir();
-    g_test_queue_destroy(unmount_dir, (char *)ns_dir);
+	g_test_queue_destroy(unmount_dir, (char *)ns_dir);
 
-    // The temporary directory should not be private initially
-    g_assert_false(sc_is_ns_group_dir_private());
+	// The temporary directory should not be private initially
+	g_assert_false(sc_is_ns_group_dir_private());
 
-    /// do what "mount --bind /foo /foo; mount --make-private /foo" does.
-    int err;
-    err = mount(ns_dir, ns_dir, NULL, MS_BIND, NULL);
-    g_assert_cmpint(err, ==, 0);
-    err = mount(NULL, ns_dir, NULL, MS_PRIVATE, NULL);
-    g_assert_cmpint(err, ==, 0);
+	/// do what "mount --bind /foo /foo; mount --make-private /foo" does.
+	int err;
+	err = mount(ns_dir, ns_dir, NULL, MS_BIND, NULL);
+	g_assert_cmpint(err, ==, 0);
+	err = mount(NULL, ns_dir, NULL, MS_PRIVATE, NULL);
+	g_assert_cmpint(err, ==, 0);
 
-    // The temporary directory should now be private
-    g_assert_true(sc_is_ns_group_dir_private());
+	// The temporary directory should now be private
+	g_assert_true(sc_is_ns_group_dir_private());
 }
-
 
 static void test_sc_initialize_ns_groups()
 {
-    if (geteuid() != 0) {
-        g_test_skip("this test needs to run as root");
-        return;
-    }
+	if (geteuid() != 0) {
+		g_test_skip("this test needs to run as root");
+		return;
+	}
 	const char *ns_dir = sc_test_use_fake_ns_dir();
-    g_test_queue_destroy(unmount_dir, (char *)ns_dir);
+	g_test_queue_destroy(unmount_dir, (char *)ns_dir);
 
-    // Initialize namespace groups using a fake directory.
-    sc_initialize_ns_groups();
+	// Initialize namespace groups using a fake directory.
+	sc_initialize_ns_groups();
 
-    // Check that the lock file did not leak unclosed.
+	// Check that the lock file did not leak unclosed.
 
 	// Construct the name of the lock file
 	char *lock_file __attribute__ ((cleanup(sc_cleanup_string))) = NULL;
-    lock_file = g_strdup_printf("%s/%s", ns_dir, SC_NS_LOCK_FILE);
-    // Attempt to open and lock the lock file.
+	lock_file = g_strdup_printf("%s/%s", ns_dir, SC_NS_LOCK_FILE);
+	// Attempt to open and lock the lock file.
 	int lock_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
 	lock_fd = open(lock_file, O_RDWR | O_CLOEXEC | O_NOFOLLOW);
 	g_assert_cmpint(lock_fd, !=, -1);
-    // The non-blocking lock operation should not fail
+	// The non-blocking lock operation should not fail
 	int err = flock(lock_fd, LOCK_EX | LOCK_NB);
 	g_assert_cmpint(err, ==, 0);
 }
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -24,7 +24,7 @@
 #include <glib.h>
 #include <glib/gstdio.h>
 
-// Set alternate namespace directory 
+// Set alternate namespace directory
 static void sc_set_ns_dir(const char *dir)
 {
 	sc_ns_dir = dir;
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -37,12 +37,23 @@ static void sc_set_ns_dir(const char *dir)
 static const char *sc_test_use_fake_ns_dir()
 {
 	char *ns_dir = NULL;
-	ns_dir = g_dir_make_tmp(NULL, NULL);
-	g_assert_nonnull(ns_dir);
+	if (g_test_subprocess()) {
+		// Check if the environment variable is set. If so then someone is already
+		// managing the temporary directory and we should not create a new one.
+		ns_dir = getenv("SNAP_CONFINE_NS_DIR");
+		g_assert_nonnull(ns_dir);
+	} else {
+		ns_dir = g_dir_make_tmp(NULL, NULL);
+		g_assert_nonnull(ns_dir);
+		g_test_queue_free(ns_dir);
+		g_assert_cmpint(setenv("SNAP_CONFINE_NS_DIR", ns_dir, 0), ==,
+				0);
+		g_test_queue_destroy((GDestroyNotify) unsetenv,
+				     "SNAP_CONFINE_NS_DIR");
+		// TODO: queue something that rm -rf's the directory tree
+	}
 	g_test_queue_destroy((GDestroyNotify) sc_set_ns_dir, SC_NS_DIR);
-	g_test_queue_free(ns_dir);
 	sc_set_ns_dir(ns_dir);
-	// TODO: queue something that rm -rf's the directory tree
 	return ns_dir;
 }
 
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -137,7 +137,7 @@ static void unmount_dir(void *dir)
     umount(dir);
 }
 
-static void test_sc_initialize_ns_groups()
+static void test_sc_is_ns_group_dir_private()
 {
     if (geteuid() != 0) {
         g_test_skip("this test needs to run as root");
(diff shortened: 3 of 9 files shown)


jdstrand:
I spent some time looking at this today and will comment inline. I can say that I may have further comments once the PR for using these functions is in place. There is a lot of 'you should only do this if that' language in the header that I think makes this module somewhat error prone to use. The header does do a good job of explaining each function though. Perhaps it should at the top also give steps on the ordering of how the functions are intended to be used?

docs/snap-confine.rst
```diff
@@ -85,6 +85,13 @@ quirks:
   the execution environment. This allows various snaps, while running in
   devmode, to access the LXD socket. LP: #1613845
 
+Sharing of the mount namespace
+------------------------------
+
+As a part of application startup `snap-confine` will move the process to a new
+mount namespace. Since version 1.0.41 all the applications belonging to a given
+snap will share the mount namespace amongst them.
```
jdstrand: Perhaps: "As of version 1.0.41 all the applications from the same snap will share the same mount namespace. Applications from different snaps continue to use separate mount namespaces."

docs/snap-confine.rst
```diff
@@ -126,6 +133,18 @@ FILES
 
 	Description of the seccomp profile.
 
+`/run/snapd/ns/.lock`:
+
+	Global lock acquired to initialize namespace sharing feature
```
jdstrand: FYI, I found the .lock file under-documented. I'm not sure where it would be best to document it, but it wasn't part of the google doc spec and I had to put it together myself why it was needed.

src/ns-support-test.c
```diff
@@ -0,0 +1,141 @@
...
+// test.
+static const char *sc_test_use_fake_ns_dir()
+{
+	char *ns_dir = NULL;
+	ns_dir = g_dir_make_tmp(NULL, NULL);
+	g_assert_nonnull(ns_dir);
+	g_test_queue_destroy((GDestroyNotify) sc_set_ns_dir, SC_NS_DIR);
+	g_test_queue_free(ns_dir);
+	sc_set_ns_dir(ns_dir);
+	// TODO: queue something that rm -rf's the directory tree
```
jdstrand: Can you implement this?

src/ns-support-test.c
```diff
@@ -0,0 +1,141 @@
...
+	group = sc_alloc_ns_group();
+	g_test_queue_free(group);
+	g_assert_nonnull(group);
+	g_assert_cmpint(group->dir_fd, ==, -1);
+	g_assert_cmpint(group->lock_fd, ==, -1);
+	g_assert_cmpint(group->event_fd, ==, -1);
+	g_assert_cmpint(group->child, ==, 0);
+	g_assert_cmpint(group->should_populate, ==, false);
+	g_assert_null(group->name);
+}
```
jdstrand, line 145: Nice test :)

src/ns-support-test.c
```diff
@@ -0,0 +1,141 @@
...
+	// file is locked by sc_nlock_ns_mutex() already.
+	int err = flock(lock_fd, LOCK_EX | LOCK_NB);
+	int saved_errno = errno;
+	g_assert_cmpint(err, ==, -1);
+	g_assert_cmpint(saved_errno, ==, EWOULDBLOCK);
+	// Unlock the namespace group mutex
+	sc_unlock_ns_mutex(group);
+	// Re-attempt the locking operation. This time it shoud succeed.
+	err = flock(lock_fd, LOCK_EX | LOCK_NB);
+	g_assert_cmpint(err, ==, 0);
```
jdstrand, line 244: For cleanliness, shouldn't we LOCK_UN after this assert?

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+#include <sched.h>
+#include <signal.h>
+#include <string.h>
+#include <sys/eventfd.h>
+#include <sys/file.h>
+#include <sys/mount.h>
+#include <sys/prctl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/types.h>
```
jdstrand: `<sys/types.h>` is specified twice.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+
+/**
+ * Directory where snap-confine keeps namespace files.
+ **/
+#define SC_NS_DIR "/run/snapd/ns"
+
+/**
+ * Effective value of SC_NS_DIR.
+ *
+ * This is only altered for testing.
```
jdstrand: Perhaps change this to "We use 'const char *' so we can update sc_ns_dir in the testsuite"

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+ * Name of the preserved mount namespace associated with SC_NS_DIR
+ * and a given group identifier (typically SNAP_NAME).
+ **/
+#define SC_NS_MNT_FILE ".mnt"
+
+// Read /proc/self/mountinfo and check if /run/snapd/ns is a private bind mount.
+//
+// That is, it cannot be shared with any other peer as defined by kernel
+// documentation listed here:
+// https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt
```
jdstrand: I think this could be worded a bit better. Eg:

```
// Read /proc/self/mountinfo and check if /run/snapd/ns is a private bind mount.
//
// We do this because /run/snapd/ns cannot be shared with any other peers as per:
// https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt
```

Also, why `//` here and `/**` up above?

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+static bool sc_is_ns_group_dir_private();
+
+void sc_initialize_ns_groups()
+{
+	int dir_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
+	debug("creating namespace group directory %s", sc_ns_dir);
+	mkpath(sc_ns_dir);
+	debug("opening namespace group directory %s", sc_ns_dir);
+	dir_fd = open(sc_ns_dir, O_DIRECTORY | O_PATH | O_CLOEXEC | O_NOFOLLOW);
+	if (dir_fd == -1) {
```
jdstrand: I tend to prefer `< 0` checks instead of `== -1`. I won't block if this stays as is. Note, you use `== -1` in quite a few places.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+// Read /proc/self/mountinfo and check if /run/snapd/ns is a private bind mount.
+//
+// That is, it cannot be shared with any other peer as defined by kernel
+// documentation listed here:
+// https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt
+static bool sc_is_ns_group_dir_private();
+
+void sc_initialize_ns_groups()
+{
+	int dir_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
```
jdstrand: Moving this down above `dir_fd = open(...` would be slightly more clear.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+		debug
+		    ("namespace group directory does not require intialization");
+	}
+	debug("unlocking the namespace group directory");
+	if (flock(lock_fd, LOCK_UN) < 0) {
+		die("cannot release lock for namespace control directory");
+	}
+}
+
+static bool sc_is_ns_group_dir_private()
```
jdstrand: I would probably prefer this to be declared above sc_initialize_ns_groups(), but I won't block on it.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+			die("cannot make the namespace group directory mount point private");
+		}
+	} else {
+		debug
+		    ("namespace group directory does not require intialization");
+	}
+	debug("unlocking the namespace group directory");
+	if (flock(lock_fd, LOCK_UN) < 0) {
+		die("cannot release lock for namespace control directory");
+	}
```
jdstrand, line 143: Perhaps a comment stating that the caller must call `sc_close_ns_group()` to close dir_fd and lock_fd would be clearer.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+#define SC_NS_MNT_FILE ".mnt"
+
+// Read /proc/self/mountinfo and check if /run/snapd/ns is a private bind mount.
+//
+// That is, it cannot be shared with any other peer as defined by kernel
+// documentation listed here:
+// https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt
+static bool sc_is_ns_group_dir_private();
+
+void sc_initialize_ns_groups()
```
jdstrand, line 102: I'm not sure there is anything to be done about it, but is seems that `sc_initialize_ns_groups()` is going to be called on every app invocation, which means every invocation will try to create `sc_ns_dir`, create the lock file and parse the mountinfo table. Since /run is tmpfs we don't have to worry about media wear and performance should be ok, plus this only has to be done when the app is launched, not for every fork/exec the app does, but thought I'd at least point it out explicitly.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	debug("opening namespace group directory %s", sc_ns_dir);
+	dir_fd = open(sc_ns_dir, O_DIRECTORY | O_PATH | O_CLOEXEC | O_NOFOLLOW);
+	if (dir_fd == -1) {
+		die("cannot open namespace group directory");
+	}
+	debug("opening lock file for group directory");
+	int lock_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
+	lock_fd = openat(dir_fd,
+			 SC_NS_LOCK_FILE,
+			 O_CREAT | O_RDWR | O_CLOEXEC | O_NOFOLLOW, 0600);
```
jdstrand: Please also add O_EXCL to this openat() call (at which point, you don't need to specify O_NOFOLLOW).

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	}
+	struct mountinfo_entry *entry = first_mountinfo_entry(info);
+	while (entry != NULL) {
+		const char *mount_dir = mountinfo_entry_mount_dir(entry);
+		const char *optional_fields =
+		    mountinfo_entry_optional_fields(entry);
+		if (strcmp(mount_dir, sc_ns_dir) == 0
+		    && strcmp(optional_fields, "") == 0) {
+			// If the /run/snapd/ns directory is mounted and has no optional
+			// fields (i.e. the shared:xxx is not there) then we are all set.
```
jdstrand: Can have more than just 'shared:xxx'. Perhaps say: `// If /run/snapd/ns has no optional fields, we know it is mounted private and there is nothing else to do`.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	if (info == NULL) {
+		die("cannot parse /proc/self/mountinfo");
+	}
+	struct mountinfo_entry *entry = first_mountinfo_entry(info);
+	while (entry != NULL) {
+		const char *mount_dir = mountinfo_entry_mount_dir(entry);
+		const char *optional_fields =
+		    mountinfo_entry_optional_fields(entry);
+		if (strcmp(mount_dir, sc_ns_dir) == 0
+		    && strcmp(optional_fields, "") == 0) {
```
jdstrand: This strcmp() is perfectly correct for today's `parse_mountinfo_entry()`, but if `parse_mountinfo_entry()` ever changes, we'll have a bug here. Can you add a comment to `parse_mountinfo_entry()` in mountinfo.c circa line 238 that we 'strcpy(entry->optional_fields, "");' to guarantee consumers have a non-null string?

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	// Name of the namespace group ($SNAP_NAME).
+	char *name;
+	// Descriptor to the namespace group control directory.  This descriptor is
+	// opened with O_PATH|O_DIRECTORY so it's only used for openat() calls.
+	int dir_fd;
+	// Descriptor to a namespace-specific lock file (i.e. $SNAP_NAME.lock).
+	int lock_fd;
+	// Descriptor to an eventfd that is used to notify the child that it can
+	// now complete its job and exit.
+	int event_fd;
```
jdstrand, line 156: I was surprised to see `event_fd` based on the spec, but I don't think it is a problem.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	// Identifier of the child process that is used during the one-time (per
+	// group) initialization and capture process.
+	pid_t child;
```
jdstrand, line 159: I know that you calloc'd this and so child is already zero, but an explicit assignment with a comment that you check for non-zero in other functions to verify the mount namespace capture process' status would make this a bit clearer.

src/ns-support.c
tyhicks, line 116: On 09/06/2016 04:17 PM, jdstrand wrote:

> In src/ns-support.c
> https://github.com/snapcore/snap-confine/pull/126#discussion_r77721186:
> 
> > +
> > +void sc_initialize_ns_groups()
> > +{
> > -    int dir_fd **attribute** ((cleanup(sc_cleanup_close))) = -1;
> > -    debug("creating namespace group directory %s", sc_ns_dir);
> > -    mkpath(sc_ns_dir);
> > -    debug("opening namespace group directory %s", sc_ns_dir);
> > -    dir_fd = open(sc_ns_dir, O_DIRECTORY | O_PATH | O_CLOEXEC | O_NOFOLLOW);
> > -    if (dir_fd == -1) {
> > -        die("cannot open namespace group directory");
> > -    }
> > -    debug("opening lock file for group directory");
> > -    int lock_fd **attribute** ((cleanup(sc_cleanup_close))) = -1;
> > -    lock_fd = openat(dir_fd,
> > -             SC_NS_LOCK_FILE,
> > -             O_CREAT | O_RDWR | O_CLOEXEC | O_NOFOLLOW, 0600);
> 
> Please also add O_EXCL to this openat() call (at which point, you don't
> need to specify O_NOFOLLOW).

I don't think that a lockfile should be opened with O_EXCL. Requiring
O_EXCL defeats the purpose of the lockfile since the exclusivity
provided by the lockfile is enforced by flock() once you've opened the file.

(Note that I have not reviewed the code in this PR at all but saw this
review comment fly by and thought I'd comment on it)

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+struct sc_ns_group *sc_open_ns_group(const char *group_name)
+{
+	struct sc_ns_group *group = sc_alloc_ns_group();
+	debug("opening namespace group directory %s", sc_ns_dir);
+	group->dir_fd =
+	    open(sc_ns_dir, O_DIRECTORY | O_PATH | O_CLOEXEC | O_NOFOLLOW);
+	if (group->dir_fd == -1) {
+		die("cannot open directory for namespace group %s", group_name);
+	}
+	char lock_fname[PATH_MAX];
```
jdstrand, line 188: I fear we may be in an 'undefined' area here. The actual full path is sc_ns_dir + '/' + lock_fname so I'm not sure what side-effects there might be if lock_name actually was PATH_MAX - 1. Really this need only be the maximum of SNAP_NAME + .lock + '\0', but I realize we don't want to be that tightly coupled with snapd's definition of how long SNAP_NAME can be. Perhaps define this length as PATH_MAX - length of sc_ns_dir + '/' + '\0'?

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	must_snprintf(lock_fname, sizeof lock_fname, "%s%s", group_name,
+		      SC_NS_LOCK_FILE);
+	debug("opening lock file for namespace group %s", group_name);
+	group->lock_fd =
+	    openat(group->dir_fd, lock_fname,
+		   O_CREAT | O_RDWR | O_CLOEXEC | O_NOFOLLOW, 0600);
```
jdstrand: Like above, can you use O_EXCL with this openat() (and optionally drop O_NOFOLLOW).

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	return group;
+}
+
+void sc_close_ns_group(struct sc_ns_group *group)
+{
+	debug("releasing resources associated wih namespace group %s",
+	      group->name);
+	close(group->dir_fd);
+	close(group->lock_fd);
+	close(group->event_fd);
```
jdstrand, line 211: Might be nice to check if dir_fd, lock_fd and event_fd are all '>= 0' here. I like to also check the return code of close(), but the dir_fd and the lock_fd aren't being written to and event_fd is only written to trigger an event, so not checking should be safe.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	free(group->name);
+	free(group);
+}
+
+void sc_lock_ns_mutex(struct sc_ns_group *group)
+{
```
jdstrand, line 217: It would be nice if `sc_lock_ns_mutex()` and `sc_unlock_ns_mutex()` both checked if lock_fd was '>=0'.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+void sc_create_or_join_ns_group(struct sc_ns_group *group)
+{
+	// Open the mount namespace file.
+	char mnt_fname[PATH_MAX];
+	must_snprintf(mnt_fname, sizeof mnt_fname, "%s%s", group->name,
+		      SC_NS_MNT_FILE);
+	int mnt_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
+	mnt_fd =
+	    openat(group->dir_fd, mnt_fname,
+		   O_CREAT | O_RDONLY | O_CLOEXEC | O_NOFOLLOW, 0600);
```
jdstrand, line 257: Can you add a comment here on why you are not using O_EXCL? AIUI, you want to either open the nsfs mounted file or open the newly created file. It would be good to document that for auditability.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	}
+	// Anything but EINVAL is an unexpected error.
+	//
+	// EINVAL is simply a sign that the file we've opened is not a valid
+	// namespace file descriptor. One potential case where this can happen is
+	// when another snap-confine tried to initialize the namespace but was
+	// killed before it managed to complete the process.
+	if (errno != EINVAL) {
+		die("cannot re-associate the mount namespace with namespace group %s", group->name);
+	}
```
jdstrand: Per the setns() man page, EINVAL might be returned for a bunch of reasons. The comment should be updated and the code reviewed to make sure the assumptions regarding EINVAL are still valid.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	debug("forking support process for mount namespace capture");
+	pid_t pid = fork();
+	debug("forked support process has pid %d", (int)pid);
+	if (pid == -1) {
+		die("cannot fork support process for mount namespace capture");
+	}
+	if (pid == 0) {
+		// This is the child process which will capture the mount namespace.
+		//
+		// It will do so by bind-mounting it the SC_NS_MNT_FILE after
```
jdstrand: `s/it//`

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+		// the parent process calls unshare() and finishes setting up the
+		// namespace completely.
+#ifdef HAVE_APPARMOR
+		// Change the hat to a sub-profile that has limited permissions
+		// necessary to accomplish the capture of the mount namespace.
+		debug
+		    ("changing apparmor hat of the support process for mount namespace capture");
+		if (aa_change_hat("mount-namespace-capture-helper", 0) < 0) {
+			die("cannot change apparmor hat of the support process for mount namespace capture");
+		}
```
jdstrand, line 314: Why `aa_change_hat()` instead of `aa_change_profile()`? This seems like a one way operation for the child so `aa_change_profile()` may be better. Also, I think you can drop `-helper` since it implies a helper binary (at least to me anyway).

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+	// The eventfd created here is used to synchronize the child and the parent
+	// processes. It effectively tells the child to perform the capture
+	// operation.
+	group->event_fd = eventfd(0, EFD_CLOEXEC);
+	if (group->event_fd == -1) {
+		die("cannot create eventfd for mount namespace capture");
+	}
+	debug("forking support process for mount namespace capture");
+	pid_t pid = fork();
+	debug("forked support process has pid %d", (int)pid);
```
jdstrand, line 297: A comment on how casting `pid` to `int` is ok with glibc would be good here.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+			die("cannot change apparmor hat of the support process for mount namespace capture");
+		}
+#endif
+		// Configure the child to die as soon as the parent dies. In an odd
+		// case where the parent is killed then we don't want to complete our
+		// task or wait for anything.
+		if (prctl(PR_SET_PDEATHSIG, SIGINT, 0, 0, 0) < 0) {
+			die("cannot set parent process death notification signal to SIGINT");
+		}
+		if (fchdir(group->dir_fd) < -1) {
```
jdstrand: This should be '< 0'.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+		// EINVAL is returned when there's nothing to unmount (no bind-mount).
+		// Instead of checking for this explicitly (which is always racy) we
+		// just unmount and check the return code.
+		if (errno != EINVAL) {
+			die("cannot unmount preserved mount namespace file %s",
+			    mnt_fname);
+		}
+	}
+	// Get back to the original directory
+	if (fchdir(old_dir_fd) < -1) {
```
jdstrand: This should be '< 0'.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+void sc_discard_preserved_ns_group(struct sc_ns_group *group)
+{
+	// Remember the current working directory
+	int old_dir_fd __attribute__ ((cleanup(sc_cleanup_close))) = -1;
+	old_dir_fd = open(".", O_PATH | O_DIRECTORY | O_CLOEXEC);
+	if (old_dir_fd == -1) {
+		die("cannot open current directory");
+	}
+	// Move to the mount namespace directory (/run/snapd/ns)
+	if (fchdir(group->dir_fd) < -1) {
```
jdstrand: This should be '< 0'.

src/ns-support.h
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * This function must be called once in each process that wishes to create or
+ * join a namespace group.
+ *
+ * It is responsible for bind mounting the control directory over itself and
+ * making it private (unsharing it with all the other peers) so that it can be
+ * used for storing preserved namespaces as bind-mounted files from the nsfs
+ * filesystem (namespace filesystem).
+ *
+ * This function acquires a floc(2)-based lock to ensure that no other instance
```
jdstrand: Typo: should be `flock`

src/ns-support.h
```diff
@@ -0,0 +1,142 @@
...
+ * Data required to manage namespaces amongst a group of processes.
+ */
+struct sc_ns_group;
+
+/**
+ * Open a namespace group.
+ *
+ * This will open and keep file descriptors for /run/snapd/ns/ as well as for
+ * /run/snapd/ns/${group_name}.lock. The lock file is created if necessary but
+ * is not locked in any way.
```
jdstrand: 'is not locked in any way' is confusing since this is a lock file. Perhaps say, 'is not locked until sc_lock_ns_mutex() is called'. Also, it seems a bit odd that the lock file is created here but not locked until another function. Perhaps document why that is?

src/ns-support.h
```diff
@@ -0,0 +1,142 @@
...
+ * @returns true if the mount namespace needs to be populated
+ **/
+void sc_create_or_join_ns_group(struct sc_ns_group *group);
+
+/**
+ * Check if the namespace needs to be populated.
+ *
+ * If the return value is true then at this stage the namespace is already
+ * unshared. The caller should perform any mount operations that are desired
+ * and then proceed to call sc_preserve_ns_group().
```
jdstrand: `sc_preserve_ns_group()` is not a thing, I think you mean `sc_preserve_populated_ns_group()`

jdstrand:
For completeness, please also add spread tests once there are consumers for `sc_initialize_ns_groups()`,  `sc_create_or_join_ns_group()`, `sc_should_populate_ns_group()`, `sc_preserve_populated_ns_group()` and `sc_discard_preserved_ns_group()`.

src/ns-support.h
```diff
@@ -0,0 +1,142 @@
...
+ * use setns() with the obtained file descriptor. If the call succeeds then the
+ * function returns and subsequent call to sc_should_populate_ns_group() will
+ * return false.
+ *
+ * If the call fails then an eventfd is constructed and a support process is
+ * forked. The child process waits until data is written to the eventfd (this
+ * can be done by calling sc_preserve_populated_ns_group()). In the meantime
+ * the parent process unshares the mount namespace() and sets a flag so that
+ * sc_should_populate_ns_group() returns true.
+ *
```
jdstrand: There is a space after this asterisk. Can you remove it?

src/ns-support.c
jdstrand: Heh, of course you are right there. :) I've deleted those two comments.

src/ns-support.c
zyga: It is easier than a pipe and more reliable (race free) than a signal handler.

src/ns-support-test.c
zyga: Yes, I just found glib lacking in performing this thing in one, nice call.

src/ns-support-test.c
zyga: Closing the file descriptor unlocks the lock. We could do it but at some point there's no need to do it.

src/ns-support.c
zyga: I moved from vm to vm and I used misconfigure vim. I'll update the docs to match.

src/ns-support.c
zyga: I realised I made this mistake just before pushing this. I was on edge between `< 0` and `-1` and I think I'll go back to `< 0` as they cannot be mistyped to `= -1` and make sense in any way.

src/ns-support.c
zyga: I'm aware of this and I was wondering if there's any way that we can avoid this without introducing some kind of raciness. I guess we could do it in a job that runs before we mount, say, the core snap, but at some point this becomes a bit distant and magic and I was worried that the essential property for being able to preserve the mount namespace would get lost.

src/ns-support.c
zyga: Yep, thanks for pointing this out

src/ns-support.c
zyga: Yes, you are right both in the deduction of the intent and the suggestion to document this better.

zyga:
Thank you for the review. It's a bit too late to do this responsibly but I will go over all the feedback and ensure it is addressed. Depending on how you feel I can propose a sister branch that uses this module to implement fully-working mount namespace sharing, along with a few spread tests for the basic property.

src/ns-support.c
```diff
@@ -0,0 +1,406 @@
...
+		if (aa_change_hat("mount-namespace-capture-helper", 0) < 0) {
+			die("cannot change apparmor hat of the support process for mount namespace capture");
+		}
+#endif
+		// Configure the child to die as soon as the parent dies. In an odd
+		// case where the parent is killed then we don't want to complete our
+		// task or wait for anything.
+		if (prctl(PR_SET_PDEATHSIG, SIGINT, 0, 0, 0) < 0) {
+			die("cannot set parent process death notification signal to SIGINT");
+		}
```
jdstrand, line 321: This may make sense to move above the aa_change_\* call, but I'm curious as to the affect on policy it will have on the parent and child profiles. I'm not terribly worried about the timing of the call-- in the unlikely event the parent dies before this call (which is not attacker controlled), the child will become a zombie and at some point init should reap it so even if the parent is killed before prctl, we should be ok.

src/ns-support.c
zyga: Fixed

src/ns-support.c
zyga: Documented

src/ns-support.c
zyga: Done, with unit tests as well

src/ns-support.c
zyga: They don't have to be >= though, e.g `event_fd` is only used if required.

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Hmm? Wait, this is not the same lock file. This is the global lock file that only this function ever touches.

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Fixed all of them to be `< 0`

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Fixed

docs/snap-confine.rst
zyga: I've improved the documentation, have a look of that is clear now.

src/ns-support-test.c
zyga: Done

docs/snap-confine.rst
zyga: Done

src/ns-support.c
zyga: Fixed

src/ns-support.c
zyga: Fixed

src/ns-support.c
zyga: Fixed

src/ns-support.h
zyga: Fixed

src/ns-support.h
zyga: Fixed

src/ns-support.c
zyga: You are right. I plan to change this a little to call `fstatfs()` to determine if the opened file descriptor belongs to `nsfs` or not. This will let us predictably die() when anything goes wrong.

src/ns-support.c
zyga: No particular reason, the profile name looked nice.

What should be the profile name here? `snap-confine//mount-namespace-capture` or just `mount-namespace-capture`?

src/ns-support.c
zyga: Done

src/ns-support.c
zyga: Let's postpone changing this until the next pull request where we actually start to use this (and we can see the apparmor profile).

src/ns-support.c
```diff
@@ -0,0 +1,425 @@
...
+	group->event_fd = eventfd(0, EFD_CLOEXEC);
+	if (group->event_fd < 0) {
+		die("cannot create eventfd for mount namespace capture");
+	}
+	debug("forking support process for mount namespace capture");
+	// Glibc defines pid as a signed 32bit integer. There's no standard way to
+	// print pid's portably so this is the best we can do.
+	pid_t pid = fork();
+	debug("forked support process has pid %d", (int)pid);
+	if (pid == -1) {
```
jdstrand: pid < 0?

src/ns-support.h
```diff
@@ -112,7 +112,7 @@ void sc_create_or_join_ns_group(struct sc_ns_group *group);
  *
  * If the return value is true then at this stage the namespace is already
  * unshared. The caller should perform any mount operations that are desired
- * and then proceed to call sc_preserve_ns_group().
+ * and then proceed to call sc_preserve_populated_ns_group().
```
jdstrand, line 115: Typo: sc_should_populate_ns_group()

src/ns-support-test.c
```diff
@@ -0,0 +1,268 @@
...
+// Set alternate namespace directory
+static void sc_set_ns_dir(const char *dir)
+{
+	sc_ns_dir = dir;
+}
+
+static void rm_rf(const char *dir)
+{
+	gint exit_status = 0;
+	gchar *cmd = g_strdup_printf("rm -rf %s", dir);
```
jdstrand: Oh, scary. I hope there are no bugs in the test code otherwise the developer might have severe data loss....

jdstrand: Can we do some checks on `dir` to make sure it is inside the test dir?

src/mountinfo.c
```diff
@@ -235,6 +235,7 @@ static struct mountinfo_entry *parse_mountinfo_entry(const char *line)
 	if ((entry->mount_opts = parse_next_string_field()) == NULL)
 		goto fail;
 	entry->optional_fields = &entry->line_buf[0] + total_used++;
+	// NOTE: This ensure s that optional_fields is never NULL.
```
jdstrand: There is a typo here. Perhaps be a little more explicit though. Eg: `// NOTE: This ensures that optional_fields is never NULL. If this changes, must adjust all callers of parse_mountinfo_entry() accordingly.`

src/ns-support-test.c
jdstrand: Alternatively remove the files we created and them rmdir the directory. If we make rmdir() fail the test, then we can catch stray files added via code changes easily. I'd prefer this approach.

src/ns-support.c
jdstrand: This sounds like a good idea.

src/ns-support.c
jdstrand: Depends on the policy. I suggest you use a child profile so it is clear this is for snap-confine. Eg:

```
@LIBEXECDIR@/snap-confine (attach_disconnected) {
    ...
    profile mount-namespace-capture (attach_disconnected) {
        ...
    }
}
```

With this you would use `@LIBEXECDIR@/snap-confine//mount-namespace-capture`.

jdstrand:
Thanks for all the changes! I think this is ok to merge provided:
- my feedback from today is addressed (ie, pid < 0, rm_rf, comment changes)
- your proposed fstatfs() is implemented

Don't feel like you need to implement the apparmor policy changes just yet as you mentioned, but let's not  cut a new version of snap-confine until the other bits are in place.

src/ns-support-test.c
zyga: I tried to use the function that doesn't do shell expansion but glib is just annoying here. I'll try again.

src/ns-support.h
zyga: Hmm?

src/ns-support.c
zyga: Fixed

src/mountinfo.c
zyga: Fixed

src/ns-support.h
jdstrand: Sorry, please disregard this. You are correct.

zyga: Resolved on IRC

src/ns-support.c
zyga: Fixed

src/ns-support.c
```diff
@@ -250,38 +252,33 @@ void sc_create_or_join_ns_group(struct sc_ns_group *group)
...
-	// namespace file descriptor. One potential case where this can happen is
-	// when another snap-confine tried to initialize the namespace but was
-	// killed before it managed to complete the process.
-	if (errno != EINVAL) {
-		die("cannot re-associate the mount namespace with namespace group %s", group->name);
-	}
-	debug
-	    ("cannot re-associate the mount namespace with namespace group %s, falling back to initialization",
-	     group->name);
+	debug("initializing new namespace group %s", group->name);
```
jdstrand, line 281: This is soo much nicer, thanks! :)

It's interesting that NSFS_MAGIC is not included in `man fstatfs`. Note that this was added only in http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/patch/include/uapi/linux/magic.h?id=e149ed2b805fefdccf7ccdfc19eca22fdd4514ac so this will likely need to be added to the backports list if it isn't already. It does seem present in the 14.04 Ubuntu kernel.

src/ns-support-test.c
```diff
@@ -251,6 +254,16 @@ static void test_sc_initialize_ns_groups()
...
 
+// Sanity check, ensure that the namespace filesystem identifier is what we
+// expect, aka NSFS_MAGIC.
+static void test_nsfs_fs_id()
+{
+	struct statfs buf;
+	int err = statfs("/proc/self/ns/mnt", &buf);
+	g_assert_cmpint(err, ==, 0);
+	g_assert_cmpint(buf.f_type, ==, NSFS_MAGIC);
+}
```
jdstrand, line 324: Good thinking! :)

src/ns-support.c
zyga: I've informed tvoss about this. With the sanity check we'll know if tests are all green. Thanks for making me aware of this :)

src/ns-support-test.c
zyga: I've implemented this a little bit better now. I think it good to merge now.

src/ns-support.c
zyga: I will follow up with a small branch that does this. Thanks

PR #127, opened 2016-09-06
Don't leak root filesystem when using pivot_root

zyga:
This patch ensures that snap-confine detaches the old root filesystem
when doing the pivot_root call. This lets the kernel trust the created
mount namespace enough to allow unprivileged user namespaces.

Fixes: https://bugs.launchpad.net/snap-confine/+bug/1618683

### DIFF
--- a/src/mount-support.c
+++ b/src/mount-support.c
@@ -210,13 +210,12 @@ static void sc_pivot_to_new_rootfs(const char *rootfs_dir)
 {
 	int old_rootfs_fd, new_rootfs_fd;
 
-	old_rootfs_fd = open("/", O_DIRECTORY | O_RDONLY | O_PATH | O_CLOEXEC);
+	old_rootfs_fd = open("/", O_DIRECTORY | O_PATH | O_CLOEXEC);
 	if (old_rootfs_fd == -1) {
 		die("cannot open old root file system directory");
 	}
 	new_rootfs_fd =
-	    open(rootfs_dir,
-		 O_DIRECTORY | O_RDONLY | O_PATH | O_NOFOLLOW | O_CLOEXEC);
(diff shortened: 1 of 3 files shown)


src/mount-support.c
```diff
@@ -193,6 +194,52 @@ static void sc_bind_mount_snap_mount_dir(const char *rootfs_dir)
...
+// chroot has now moved all paths to /tmp/snap.rootfs_*.
+//
+// Because we are using unshare(2) with CLONE_NEWNS we can essentially use
+// pivot_root just like chroot but this makes apparmor unaware of the old root
+// so everything works okay.
+static void sc_pivot_to_new_rootfs(const char *rootfs_dir)
+{
+	int old_rootfs_fd, new_rootfs_fd;
+
+	old_rootfs_fd = open("/", O_DIRECTORY | O_RDONLY | O_PATH | O_CLOEXEC);
```
stgraber: O_RDONLY is meaningless here as it's ignored by the kernel when O_PATH is passed.

src/mount-support.c
```diff
@@ -193,6 +194,52 @@ static void sc_bind_mount_snap_mount_dir(const char *rootfs_dir)
...
+	if (old_rootfs_fd == -1) {
+		die("cannot open old root file system directory");
+	}
+	new_rootfs_fd =
+	    open(rootfs_dir,
+		 O_DIRECTORY | O_RDONLY | O_PATH | O_NOFOLLOW | O_CLOEXEC);
```
stgraber: Same thing, you can remove O_RDONLY from there.

zyga: Ah, right, thanks for spotting that.

src/mount-support.c
zyga: Fixed

PR #128, opened 2016-09-08
allow read/write to /dev/pts/[0-9]* needed for running snaps under a confined sshd

jdstrand:


zyga:
FYI:

```
22:12 < jdstrand> zyga: hey, while you are in there, I noticed that snap-confine needs this when running over ssh: /dev/pts/[0-9]* rw,
22:13 < jdstrand> apparmor="DENIED" operation="file_inherit" profile="/usr/lib/snapd/snap-confine" name="/dev/pts/2" pid=28375 comm="ubuntu-core-lau" 
                  requested_mask="wr" denied_mask="wr" fsuid=1000 ouid=1000
22:13 < jdstrand> zyga: I noticed this ssh'ing into a xenial classic system and using 'hello-world'
22:14 < jdstrand> zyga: I can do a PR if it is easier
22:26 < jdstrand> zyga: hoping you are EOD. I'll do a PR
22:31 < jdstrand> zyga: fyi, https://github.com/snapcore/snap-confine/pull/128
22:37 < zyga> jdstrand: hmm, curious why is that?
22:37 < zyga> jdstrand: I work over ssh all day
22:37 < zyga> jdstrand: and I didn't need it
```

jdstrand:
I forgot-- on this system sshd is confined and I'm using pam-apparmor. The inherited fd is mediated since it isn't coming from an unconfined process.

zyga:
Looks good, thank you

PR #134, opened 2016-09-12
 Enable snap-confine namespace sharing

zyga:
This branch changes mount-support.[ch] a little so that there's no
explicit unshare API anymore (this is handled by ns-support.h) and so
that mount-suppor.h is really just about populating the namespace that
is already provided.

In addition, sc-main.c now uses SNAP_NAME to join or create a namespace
group and if populates it if necessary. The apparmor profile is updated
to let snap-confine perform the additional tasks.


zyga:
I'll close this for now and reopen once #135 lands

zyga:
I'll close this until all the tests run in qemu

PR #135, opened 2016-09-12
Add snap-discard-ns

zyga:
This branch adds the `snap-discard-ns` program that does as the name suggests. It is a part of the namespace sharing feature. This program will be used by snap-confine spread tests before the appropriate branch lands in snapd (https://github.com/snapcore/snapd/pull/1847)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 src/snap-confine
+src/snap-discard-ns
 src/snap-confine-unit-tests
 src/snap-confine.apparmor
 *~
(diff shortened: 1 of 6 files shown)


tyhicks:
This should all be good as long as /run/snapd/ and /run/snapd/ns/ is only writable by root. That's true for /run/snapd/ns/ because sc_initialize_ns_groups() calls mkpath() which does the right thing by creating the directory with the 0755 mode.

snapd must create /run/snapd/. Can you verify that snapd is creating that directory with the right permissions?

tyhicks:
@zyga and I discussed my last question in #snappy. He informed me that /run/snapd/ is created by sc_initialize_ns_groups() and we already know that it creates dirs with 0755 mode.

Looks good to me. Thanks!

PR #136, opened 2016-09-13
Add support for running spread tests in qemu

zyga:
This branch adds support for spread tests that can be ran locally with qemu.

As an optimization, the environment variable APT_PROXY can be pointed to a local apt-cacher-ng instance to optimize package downloads.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -27,7 +27,7 @@ exclude:
 
 prepare: |
     echo "Spread is running as $(id)"
-    [ "$REUSE_PROJECT" != 1 ] || exit 0
+    [ "${REUSE_PROJECT:-}" != 1 ] || exit 0
     ./spread-tests/spread-prepare.sh
 
 suites:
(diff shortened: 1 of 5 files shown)


spread.yaml
```diff
@@ -1,15 +1,22 @@
...
 
 backends:
     linode:
         key: "$(HOST: echo $SPREAD_LINODE_KEY)"
         systems:
             - ubuntu-16.04-64-grub
             # - ubuntu-16.04-32-grub
+    qemu:
+        systems:
+            - ubuntu-16.04:
```
mvo5, line 15: you may want to add ubuntu-16.10 here, we will need it at some point :)

mvo5:
Thanks, looks very nice!

spread.yaml
zyga: I tried this but apparently `adt-buildvm-ubuntu-cloud` cannot yet built an image for it.

PR #137, opened 2016-09-13
chdir to original directory after setns

zyga:
This patch fixes snap-confine's current working directory after a
successful setns call. This apparently a little undocumented but setns
coupled with CLONE_NEWNS resets the current working directory to the
root directory.

This patch is a quick reaction, I will refactor all cwd management code
with a separate branch as currently too many things need to care about
it.

### DIFF
--- a/src/ns-support.c
+++ b/src/ns-support.c
@@ -45,6 +45,14 @@
 #include "mountinfo.h"
 #include "cleanup-funcs.h"
 
+/*!
+ * The void directory.
+ *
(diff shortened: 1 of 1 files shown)


chipaca:
:+1:

jdstrand:
This patch looks fine though where is /var/lib/snapd/void created?

zyga:
@jdstrand it is created by `make install` (it is also appropriately managed by packaging)

PR #138, opened 2016-09-13
Discard snapd-hacker-toolbelt namespaces on restore

zyga:
Because tests assume pristine test environment and preserved mount
namespaces are not yet discarded by snapd on the removal of a given snap
(branch pending but not merged) we just do it by hand.

Since not all of the tests currently use the snapd-hacker-toolbelt snap
the cleanup code is allowed to fail by ignoring any errors.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -33,5 +33,12 @@ prepare: |
 suites:
     spread-tests/main/:
         summary: Full-system tests for snap-confine
+        restore-each: |
+            # NOTE: before snapd discards namespaces on snap removal
+            # we have to do it ourselves. All of the tests use one
+            # snap so this is easier to work with.
+            /usr/lib/snapd/snap-discard-ns snapd-hacker-toolbelt || :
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -33,5 +33,12 @@ prepare: |
 suites:
     spread-tests/main/:
         summary: Full-system tests for snap-confine
+        restore-each: |
+            # NOTE: before snapd discards namespaces on snap removal
+            # we have to do it ourselves. All of the tests use one
+            # snap so this is easier to work with.
+            /usr/lib/snapd/snap-discard-ns snapd-hacker-toolbelt || :
```
chipaca, line 40: fwiw I find `|| true` to be a little less cryptic.

chipaca:
:+1:

spread.yaml
zyga: Old habits die hard, I'll change this to `true`

PR #139, opened 2016-09-13
Allow snap-confine to read the void directory

zyga:
The apparmor profile for snap-confine needs to allow reading of the
/var/lib/snapd/void directory so that we can safely chdir there.


jdstrand:
On Tuesday, September 13, 2016 3:35:27 PM CDT, Zygmunt Krynicki 
notifications@github.com wrote:

> The apparmor profile for snap-confine needs to allow reading of the
> /var/lib/snapd/void directory so that we can safely chdir there.
> 
> Signed-off-by: Zygmunt Krynicki zygmunt.krynicki@canonical.com

+1

## 

Sent using Dekko from my Ubuntu device

PR #140, opened 2016-09-14
Fallback to getenv when secure_ version is missing

lool:
Function secure_getenv is a GNU extension; fallback to getenv on systems which lack secure_getenv, but warn during configure. This is useful to non-glibc ports such as OpenWRT which uses musl libc by default.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -161,5 +161,15 @@ AS_IF([test "x$enable_caps_over_setuid" = "xyes"], [
     AC_DEFINE([CAPS_OVER_SETUID], [1],
         [Use capabilities rather than setuid bit])])
 
+AC_CHECK_FUNCS([secure_getenv], [], [
+    AC_MSG_WARN([
+    XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+    X                                              X
+    X secure_getenv missing; suid binaries at risk X
+    X                                              X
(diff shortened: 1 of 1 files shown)


zyga:
I wonder how hard would it be to implement secure_getenv from scratch. All you need to do is to look at AT_SECURE in the auxiliary vector and return NULL if it is set. Have a look at http://man7.org/linux/man-pages/man3/getauxval.3.html for details.

zyga:
Loic would you mind reporting a bug on launchpad.net/snap-confine so that I can track it as belonging to one of the milestones.

jdstrand:
Note that there are quite a few gnu extensions used in snap-confine.

jdstrand:
I like @zyga's idea of reimplementing secure_getenv(), but can we make sure we use glibc's version when it is available and the reimplementation when it is not?

lool:
I filed https://bugs.launchpad.net/snap-confine/+bug/1623725

I did consider providing a secure_getenv impl as a fallback, but given that we disable apparmor/seccomp on openwrt anyway, I didn't bother.

jdstrand:
NAK - two members of the security team advised reimplementing secure_getenv. Please undo this commit. Disabling other security backends shouldn't be a justification for reimplementing secure_getenv.

zyga:
I've implemented a secure_getenv version and I'll use it instead.

PR #141, opened 2016-09-14
Ensure that parent is alive after installing PR_SET_PDEATHSIG

zyga:
The child process used for mount namespace capture uses
prctl(PR_SET_PDEATHSIG, ...) to get a SIGINT signal when the parent
process dies. The problem is that by the time that code runs the parent
process may have already died and the signal will never arrive.

To avoid this ensure that the parent process is still alive with
kill(parent, 0). The parent is not interrogated with getppid() but
instead is inherited from the original parent across fork as the child
process is re-parented to the init process (or locally installed
subreaper) when that occurs.

### DIFF
--- a/src/ns-support.c
+++ b/src/ns-support.c
@@ -350,8 +350,10 @@ void sc_create_or_join_ns_group(struct sc_ns_group *group)
 			die("cannot set parent process death notification signal to SIGINT");
 		}
 		// Check that parent process is still alive. If this is the case then
-		// we can reliably rely on the PR_SET_PDEATHSIG signal to wake us up
-		// from eventfd_read() below.
+		// we can *almost* reliably rely on the PR_SET_PDEATHSIG signal to wake
+		// us up from eventfd_read() below. In the rare case that the PID numbers
+		// overflow and the now-dead parent PID is recycled we will still hang
+		// forever on the read from eventfd below.
 		debug("ensuring that parent process is still alive");
 		if (kill(parent, 0) < 0) {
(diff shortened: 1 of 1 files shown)


src/ns-support.c
```diff
@@ -345,6 +349,20 @@ void sc_create_or_join_ns_group(struct sc_ns_group *group)
...
+		if (kill(parent, 0) < 0) {
+			switch (errno) {
+			case ESRCH:
+				debug("parent process has already terminated");
+				abort();
+			default:
+				die("cannot ensure that parent process is still alive");
+				break;
+			}
+		}
```
jdstrand, line 367: This is definitely better than no check, but this check is not perfect. The parent could die and the kernel could assign the parent pid to a new process. Perhaps it would be better to simply timeout. The parent is supposed to send that signal quite fast... if the child doesn't get that in say, 10 seconds, just debug("parent process did not signal with 10 seconds"); abort();

zyga: I have this in a separate branch and I agree we should do both.

jdstrand: After talking on IRC, the intent of this is simply an early warning system if the parent dies right away and that an in progress PR will implement a timeout. In this light, this PR is fine to land as is-- in the normal case the pid won't be reused and we can bail fast but if the pid is reused, that process won't send the signal anyway and the timeout will expire and we can cleanup then.

+1 provided that a comment is added regarding pid reuse and that we'll use a timeout to handle that.

PR #142, opened 2016-09-14
Add sanity timeouts

zyga:
This patch adds library functions to ensure that ensures a system call
such as flock() doesn't block for more than a given "sanity timeout"
value. This is meant to guard against bugs in the code that might
manifest as hanging process, keeping a flock-based lock alive, that will
never wake up.

The timer is implemented using SIGARLM and alarm(). The intent is to
simply wake up the system call and detect a flag being set by the signal
handler. This relies on using a signal handler that is not restarting
system calls.

### DIFF
--- a/src/ns-support.c
+++ b/src/ns-support.c
@@ -67,7 +67,8 @@ static void sc_SIGALRM_handler(int signum)
  * are interrupted and a flag is set.
  *
  * The call should be paired with sc_disable_sanity_check_timeout() that
- * disabled the alarm and acts on the flag, aborting the process.
+ * disables the alarm and acts on the flag, aborting the process if the timeout
+ * gets exceeded.
  **/
 static void __attribute__ ((used)) sc_enable_sanity_timeout()
 {
--- a/src/ns-support.c
+++ b/src/ns-support.c
@@ -91,13 +91,13 @@ static void __attribute__ ((used)) sc_enable_sanity_timeout()
  **/
 static void __attribute__ ((used)) sc_disable_sanity_timeout()
 {
+	if (sanity_timeout_expired) {
+		die("sanity timeout expired");
+	}
 	alarm(0);
 	struct sigaction act = { };
 	if (sigaction(SIGALRM, &act, NULL) < 0) {
 		die("cannot uninstall signal handler for SIGALRM");
-	}
-	if (sanity_timeout_expired) {
-		die("sanity timeout expired");
 	}
 	debug("sanity timeout reset and disabled");
 }
--- a/src/ns-support.c
+++ b/src/ns-support.c
@@ -45,6 +45,63 @@
 #include "mountinfo.h"
 #include "cleanup-funcs.h"
 
+/**
+ * Flag indicating that a sanity timeout has expired.
+ **/
+static volatile sig_atomic_t sanity_timeout_expired = 0;
+
+/**
+ * Signal handler for SIGALRM that sets sanity_timeout_expired flag to 1.
+ **/
+static void sc_SIGALRM_handler(int signum)
+{
+	sanity_timeout_expired = 1;
+}
+
+/**
+ * Enable a sanity-check timeout.
+ *
+ * The timeout is based on good-old alarm(2) and is intended to break a
+ * suspended system call, such as flock, after a few seconds. The built-in
+ * timeout is primed for three seconds. After that any sleeping system calls
+ * are interrupted and a flag is set.
+ *
+ * The call should be paired with sc_disable_sanity_check_timeout() that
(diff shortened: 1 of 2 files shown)


src/ns-support.c
```diff
@@ -45,6 +45,63 @@
...
+ **/
+static void sc_disable_sanity_timeout()
+{
+	alarm(0);
+	debug("reset and disabled the failure timer");
+	struct sigaction act = { };
+	if (sigaction(SIGALRM, &act, NULL) < 0) {
+		die("cannot uninstall signal handler for SIGALRM");
+	}
+	if (sanity_timeout_expired) {
```
zyga: I guess this could be done before we reset the alarm really

src/ns-support.c
```diff
@@ -45,6 +45,63 @@
...
+static void sc_enable_sanity_timeout()
+{
+	sanity_timeout_expired = 0;
+	struct sigaction act = {.sa_handler = sc_SIGALRM_handler };
+	// NOTE: we are using sigaction so that we can explicitly control signal
+	// flags and *not* pass the SA_RESTART flag. The intent is so that any
+	// system call we may be sleeping on to get interrupted.
+	if (sigaction(SIGALRM, &act, NULL) < 0) {
+		die("cannot install signal handler for SIGALRM");
+	}
```
jdstrand, line 86: This is more standard (eg, http://www.gnu.org/software/libc/manual/html_node/Sigaction-Function-Example.html):

```
struct sigaction act;
act.sa_handler = sc_SIGALRM_handler;
if (sigemptyset(&act.sa_mask) < 0) {
    die(...);
}
act.sa_flags = 0;
if (sigaction(SIGINT, &act, NULL) < 0) {
    die(...);
}
```

If you use this, you can drop the comment since it is clear the flag won't be set.

src/ns-support.c
```diff
@@ -45,6 +45,63 @@
...
+static void __attribute__ ((used)) sc_disable_sanity_timeout()
+{
+	if (sanity_timeout_expired) {
+		die("sanity timeout expired");
+	}
+	alarm(0);
+	struct sigaction act = { };
+	if (sigaction(SIGALRM, &act, NULL) < 0) {
+		die("cannot uninstall signal handler for SIGALRM");
+	}
```
jdstrand, line 109: Similarly, I think you would want to use this instead:

```
struct sigaction act;
if (sigemptyset(&act.sa_mask) < 0) {
    die(...);
}
if (sigaction(SIGINT, &act, NULL) < 0) {
    die(...);
}
```

However, rather than setting this to the empty set, perhaps it would be better to save off the old action in sc_enable_sanity_timeout() and then reinstating it here. The GNU libc URL above shows how to do this; I suspect that may be overkill.

src/ns-support.c
zyga: Done, slightly differently, have a look please.

src/ns-support.c
zyga: I assume you meant `SIGARLM`? I also tweaked this to restore `SIG_DFL`. I don't think we need to restore anything else as we're not using any other signals at this time (YAGNI)

src/ns-support.c
```diff
@@ -45,6 +45,69 @@
...
+ * The call should be paired with sc_disable_sanity_check_timeout() that
+ * disables the alarm and acts on the flag, aborting the process if the timeout
+ * gets exceeded.
+ **/
+static void __attribute__ ((used)) sc_enable_sanity_timeout()
+{
+	sanity_timeout_expired = 0;
+	struct sigaction act = {.sa_handler = sc_SIGALRM_handler };
+	if (sigemptyset(&act.sa_mask) < 0)
+		die("cannot initialize POSIX signal set");
```
jdstrand, line 78: Nitpick: you aren't using '{}' for a one line body here but you do a few lines below. Does this pass indent?

src/ns-support.c
```diff
@@ -45,6 +45,69 @@
...
+ **/
+static void __attribute__ ((used)) sc_disable_sanity_timeout()
+{
+	if (sanity_timeout_expired) {
+		die("sanity timeout expired");
+	}
+	alarm(0);
+	struct sigaction act = {.sa_handler = SIG_DFL };
+	if (sigemptyset(&act.sa_mask) < 0)
+		die("cannot initialize POSIX signal set");
```
jdstrand, line 105: Same comment here re '{}'

jdstrand:
+1 provided my style comments are addressed.

src/ns-support.c
zyga: It does, I'll correct this to { }

PR #143, opened 2016-09-14
Don't assume /run/snapd/ns can be removed

zyga:
This patch fixes a test case that is overzealous in the assumption that
/run/snapd/ns can be removed. That directory can contain additional
mount and lock files from any snaps present on the system that have run
earlier.

### DIFF
--- a/spread-tests/main/discard-ns/task.yaml
+++ b/spread-tests/main/discard-ns/task.yaml
@@ -18,3 +18,6 @@ restore: |
     umount /run/snapd/ns
     rm /run/snapd/ns/foo.mnt
     rm /run/snapd/ns/foo.lock
+    # The removal is optional as the directory may contain other files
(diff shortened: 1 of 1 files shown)


spread-tests/main/discard-ns/task.yaml
```diff
@@ -18,4 +18,3 @@ restore: |
     umount /run/snapd/ns
     rm /run/snapd/ns/foo.mnt
     rm /run/snapd/ns/foo.lock
-    rmdir /run/snapd/ns
```
jdstrand, line 21: What about "rmdir /run/snapd/ns || true" instead (with a comment that it is ok when it can't be removed). This way we get the cleanup if it can be done.

zyga: Done

PR #144, opened 2016-09-14
Ensure that snap-confine is dead after each test terminates

zyga:
This is just a sanity check so that no copy of snap-confine is around
after test terminates. If something ever breaks where the child process
may end up still hanging we will know.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -34,6 +34,9 @@ suites:
     spread-tests/main/:
         summary: Full-system tests for snap-confine
         restore-each: |
+            # Ensure that we don't have stranded processes around
+            test "$(pgrep ubuntu-core-launcher | wc -l)" -eq 0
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -34,6 +34,9 @@ suites:
     spread-tests/main/:
         summary: Full-system tests for snap-confine
         restore-each: |
+            # Ensure that we don't have stranded processes around
+            test "$(pgrep ubuntu-core-launcher | wc -l)" -eq 0
+            test "$(pgrep snap-confine | wc -l)" -eq 0
```
jdstrand: You can use 'pgrep -c' instead of piping to `wc -l`.

+1 with that change.

zyga: Done, thanks for the suggestion. I didn't know about it

PR #146, opened 2016-09-14
Use sanity timeouts around blocking operations

zyga:
This patch uses the sanity timeout mechanism around both locking
flock(2) calls and around the eventfd_read(2) call. This ensures that
even if the call blocks due to unknown bugs in the code, snap-confine
will die at within a predictable amount of time.


jdstrand:
+1

PR #147, opened 2016-09-19
Implement secure_getenv(3) if not provided by stdlib

zyga:
This patch changes the previous approach to falling back to unsecure
getenv if the standard library (e.g. musl) didn't provide the
secure_getenv() function. Snap confine now provides an internal copy of
the required library function instead.

### DIFF
--- a/src/secure-getenv.c
+++ b/src/secure-getenv.c
@@ -23,8 +23,9 @@
 char *secure_getenv(const char *name)
 {
 	unsigned long secure = getauxval(AT_SECURE);
-	if (secure != 0)
+	if (secure != 0) {
 		return NULL;
+	}
 	return getenv(name);
 }
 #endif				// ! HAVE_SECURE_GETENV
--- a/src/secure-getenv.c
+++ b/src/secure-getenv.c
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2016 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


src/secure-getenv.c
```diff
@@ -0,0 +1,30 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#include "secure-getenv.h"
+
+#include <stdlib.h>
+#include <sys/auxv.h>
+
+#ifndef HAVE_SECURE_GETENV
+char *secure_getenv(const char *name)
```
chipaca, line 23: you could add `__attribute__((nonnull))` so that gcc and clang warn you if you pass in null, i think?

src/secure-getenv.c
```diff
@@ -0,0 +1,30 @@
...
+{
+	unsigned long secure = getauxval(AT_SECURE);
+	if (secure != 0)
```
chipaca: always use braces around ifs that aren't single-liners? Or maybe just always use braces around ifs.

zyga:
Spread failures are caused by out-of-date things in the image. Let me try to get spread to update everything first.

src/secure-getenv.c
chipaca: also `warn_unused_result` fwiw

chipaca (approved):
+1 with a nit and some suggestions

src/secure-getenv.c
zyga: Nice, I'll do it

src/secure-getenv.c
zyga: { } are like seat belts, will do

zyga (commented):
Thanks for the review. I'll update the pull request.

lool:
Looks good to me too, thanks for adding this

PR #150, opened 2016-09-20
Skip tests that require Linux kernel 3.19+

zyga:
The test sanity code that checks if /proc/self/ns/mnt is an NSFS_MAGIC
filesystem is skipped if invoked on a Kernel older than 3.19. This lets
the package build in an older kernel and a newer chroot sensibly (e.g.
in a launchpad builder).


jdstrand:
+1

PR #151, opened 2016-09-22
Make snap-discard-ns fail gracefully

zyga:
This patch changes snap-discard-ns to gracefully fail (as in, return
successfully) when any of the following three conditions happen:
- the /run/snapd/ns directory doesn't exist yet
- the /run/snapd/ns/$SNAP_NAME.mnt file doesn't exist
- the /run/snapd/ns/$SNAP_NAME.mnt file is not a mount point

This allows snapd to check for abnormal errors from snap-discard-ns.

### DIFF
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -155,7 +155,7 @@ static struct sc_ns_group *sc_test_open_ns_group(const char *group_name)
 	if (group_name == NULL) {
 		group_name = "test-group";
 	}
-	group = sc_open_ns_group(group_name);
+	group = sc_open_ns_group(group_name, 0);
 	g_test_queue_destroy((GDestroyNotify) sc_close_ns_group, group);
 	// Check if the returned group data looks okay
 	g_assert_nonnull(group);
--- a/src/ns-support-test.c
+++ b/src/ns-support-test.c
@@ -185,6 +185,15 @@ static void test_sc_open_ns_group()
 		      (lock_file, G_FILE_TEST_EXISTS | G_FILE_TEST_IS_REGULAR));
 }
 
+static void test_sc_open_ns_group_graceful()
+{
+	sc_set_ns_dir("/nonexistent");
+	g_test_queue_destroy((GDestroyNotify) sc_set_ns_dir, SC_NS_DIR);
+	struct sc_ns_group *group =
+	    sc_open_ns_group("foo", SC_NS_FAIL_GRACEFULLY);
+	g_assert_null(group);
+}
+
 static void test_sc_lock_ns_mutex_precondition()
 {
 	sc_test_use_fake_ns_dir();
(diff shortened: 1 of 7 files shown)


src/ns-support.c
```diff
@@ -277,6 +277,36 @@ struct sc_ns_group *sc_open_ns_group(const char *group_name)
...
+{
+	struct sc_ns_group *group = sc_alloc_ns_group();
+	debug("opening namespace group directory %s", sc_ns_dir);
+	group->dir_fd =
+	    open(sc_ns_dir, O_DIRECTORY | O_PATH | O_CLOEXEC | O_NOFOLLOW);
+	if (group->dir_fd < 0) {
+		if (errno == ENOENT) {
+			free(group);
+			return NULL;
+		}
```
jdstrand, line 265: This function is identical to sc_open_ns_group() except for this ENOENT check. It would be more maintainable to instead modify sc_open_ns_group() to take a flag and decide how to error based on the flag. Eg:

```
struct sc_ns_group *sc_open_ns_group(const char *group_name, const int graceful_enoent)
{
...
    if (group->dir_fd < 0) {
                if (graceful_enoent == 1 && errno == ENOENT) {
                        free(group);
                        return NULL;
                }
                die("cannot open directory for namespace group %s", group_name);
        }
...
```

Then sc-main.c uses:

```
group = sc_open_ns_group(group_name, 0);
```

and snap-discard-ns.c uses:

```
group = sc_open_ns_group(group_name, 1);
```

Obviously feel free to rename graceful_enoent and be sure to document this argument.

jdstrand (changes requested):
Please adjust sc_open_ns_group() to take an arg instead as per inline comments

src/ns-support.c
zyga: Will do, thank you for the review

zyga:
Spread failure looks related to packaging. I will fix this separately and improve spread setup so that we see what actually failed.

src/ns-support-test.c
```diff
@@ -155,7 +155,7 @@ static struct sc_ns_group *sc_test_open_ns_group(const char *group_name)
 	if (group_name == NULL) {
 		group_name = "test-group";
 	}
-	group = sc_open_ns_group(group_name);
+	group = sc_open_ns_group(group_name, 0);
 	g_test_queue_destroy((GDestroyNotify) sc_close_ns_group, group);
 	// Check if the returned group data looks okay
 	g_assert_nonnull(group);
```
jdstrand, line 161: +1 provided you add a test for `sc_open_ns_group(group_name, SC_NS_FAIL_GRACEFULLY)`

zyga: Will do, thank you! :)

jdstrand (approved):

zyga:
I'm merging this because the test failures are not really related to the change being made. Analysis indicates that ubuntu-core is being unmounted at some (random) point during the test process and depending on how fast or slow the VM is, tests just randomly fail when that happens. Snap-confine has no code that performs sanity checks on the presence of the core snap.

PR #153, opened 2016-09-23
add clone(CLONE_NEWUSER) test in spread-tests/regression/lp-1618683b

jdstrand:
There is already a test for 'unshare -U' in spread-tests/regression/lp-1618683, but the fix for LP: #1618683 also fixed clone(CLONE_NEWUSER), so let's add a test for that. This requires merging https://code.launchpad.net/~jdstrand/+git/snapd-hacker-toolbelt/+ref/master.

### DIFF
--- a/spread-tests/regression/lp-1618683b/task.yaml
+++ b/spread-tests/regression/lp-1618683b/task.yaml
@@ -1,0 +1,16 @@
+summary: Check that CLONE_NEWUSER works with clone() within snap apps
+# This is blacklisted on debian because we first have to get the dpkg-vendor patches
+systems: [-debian-8]
+details: |
+    Snap-confine used to "leak" the root filesystem directory across the
+    pivot_root call. This caused checks in the kernel to fail and resulted in
+    the inability to use clone(CLONE_NEWUSER) from snaps.
+prepare: |
+    echo "Having installed the snapd-hacker-toolbelt snap"
+    snap install snapd-hacker-toolbelt
+execute: |
+    cd /
+    echo "We can clone(CLONE_NEWUSER) as a regular user and expect it to work"
+    /snap/bin/snapd-hacker-toolbelt.busybox sh -c 'clone-newuser-test'
+restore: |
+    snap remove snapd-hacker-toolbelt


zyga (commented):
There's an existing test that does this so I'm interested in knowing if we need both. Have a look at https://github.com/snapcore/snap-confine/blob/master/spread-tests/regression/lp-1618683/task.yaml

spread-tests/regression/lp-1618683b/task.yaml
```diff
@@ -0,0 +1,16 @@
...
+    Snap-confine used to "leak" the root filesystem directory across the
+    pivot_root call. This caused checks in the kernel to fail and resulted in
+    the inability to use clone(CLONE_NEWUSER) from snaps.
+prepare: |
+    echo "Having installed the snapd-hacker-toolbelt snap"
+    snap install snapd-hacker-toolbelt
+execute: |
+    cd /
+    echo "We can clone(CLONE_NEWUSER) as a regular user and expect it to work"
+    /snap/bin/snapd-hacker-toolbelt.busybox sh -c 'clone-newuser-test'
```
zyga, line 14: Can you tell me what `clone-newuser-test` is and where does it come from?

jdstrand:
"There's an existing test that does this so I'm interested in knowing if we need both. Have a look at https://github.com/snapcore/snap-confine/blob/master/spread-tests/regression/lp-1618683/task.yaml"

I thought I described this in the PR...: "There is already a test for 'unshare -U' in spread-tests/regression/lp-1618683, but the fix for LP: #1618683 also fixed clone(CLONE_NEWUSER), so let's add a test for that."

Your unshare test specifies a devmode snap. The test in the PR works in strict mode. Could unshare be replaced by this test? Possibly, but testing in both devmode and strict mode seemed worthwhile.

spread-tests/regression/lp-1618683b/task.yaml
jdstrand: "Can you tell me what clone-newuser-test is and where does it come from?"

I addressed this in the PR: "This requires merging https://code.launchpad.net/~jdstrand/+git/snapd-hacker-toolbelt/+ref/master."

If you are wondering where that came from-- it came from me-- I wrote clone-newuser-test.c.

zyga:
Thank you for the explanation, it makes sense now.

zyga:
I'll merge this for the next release. Thank you for writing this.

PR #155, opened 2016-09-26
add README.mount_namespace to provide detailed information on mount setup

jdstrand:
This document provides the information necessary to create straces of
snap-confine and details the pertinent syscalls to tersely describe the
complicated mount namespace setup and how all snaps and classic differ. This
should be useful for anyone debugging snap-confine, code reviewers, auditors,
etc.

### DIFF
--- a/README.mount_namespace
+++ b/README.mount_namespace
@@ -1,13 +1,24 @@
 = Mount namespace setup in snap-confine =
 
+This document provides a terse explanation of the mount setup using syscall
+traces to show precisely what is happening and show the difference between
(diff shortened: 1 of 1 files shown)


jdstrand:
The test failure is unrelated to this change:

```
2016/09/26 22:31:17 Cannot allocate linode:ubuntu-16.04-64-grub: authentication failed
```

zyga (approved):
Thank you for writing this. I somewhat worry about the cost of ongoing maintenance but I think this is too valuable not to merge

PR #156, opened 2016-09-27
Check the real alternatives directory in core_snap_dir

binli:
We found an issue when run the snapd in BQ tablet device, it will bind /snap/ubuntu-core/current/etc/alternatives, but before that we don't check the user's permissions with access.

### DIFF
--- a/src/mount-support.c
+++ b/src/mount-support.c
@@ -312,11 +312,15 @@ static void setup_snappy_os_mounts()
 			      etc_alternatives);
 		must_snprintf(dst, sizeof dst, "%s%s", rootfs_dir,
 			      etc_alternatives);
-		debug("bind mounting %s to %s", src, dst);
-		// NOTE: MS_SLAVE so that the started process cannot maliciously mount
-		// anything into those places and affect the system on the outside.
-		if (mount(src, dst, NULL, MS_BIND | MS_SLAVE, NULL) != 0) {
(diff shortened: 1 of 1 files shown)


zyga (commented):
I'm not quite sure I understand. Snap-confine runs as root so there's no permission check that is required. Can you tell me what the problem was in more detail?

binli:
@zyga 

 Thanks for your reply, I tried to enable the snapd in tablet, there are no systemd, I tried to use the snapd-wrapper from link[1]. When I try to run hello-world, it will reboot the system, after check, it will hang when do mount bind the /snap/ubuntu-core/current/etc/alternatives, but in code it don't check this directory's permission, it check the /etc/alternatives, but we don't use it, after I check the permission of correct directory with access, it could run hello-world successfully. :) 

[1] https://github.com/teknoraver/snap-openwrt.git

zyga:
Hmm, but if snap-confine runs as root (it has to) then the permission check makes no sense. I must be missing something. Let me tell you this, this code is going away this week (a major change is going to land soon) so Iet's wait until next week to see how it would look like (similar but different code will exist). It would also help if you could ping me next week to explain how to reproduce this (I have a BQ 4.5E if that helps)

binli:
zyga,

 Now we enabled the systemd to replace upstart, after that I tried again, I found this issue dismissed, :D.

PR #158, opened 2016-09-28
Constrain SNAP_CONFINE_DEBUG values

zyga:
This patch changes how the aforementioned environment variable is
interpreted. To have some sanity only a subset of values now triggers
the debugging semantics. Specifically only "yes" and "1" do so, other
values are handled gracefully (e.g. unset, empty, "no" or "0") and
finally all other values result in a diagnostic message.

### DIFF
--- a/src/utils-test.c
+++ b/src/utils-test.c
@@ -17,3 +17,48 @@
 
 #include "utils.h"
 #include "utils.c"
+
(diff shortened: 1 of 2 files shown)


Conan-Kudo (approved):
Looks good to me!

PR #159, opened 2016-09-28
Add mmap for nested containers on 4.8

jdstrand:
4.8+ kernels have a semantic change where the location of the mmap check in
the binfmt_elf loader changed along with the cred that is used for the check.
As a result, when using snap-confine in an LXD container that supports AppArmor
namespace stacking we must allow 'm' on @LIBEXECDIR@/snap-confine.

### DIFF
--- a/src/snap-confine.apparmor.in
+++ b/src/snap-confine.apparmor.in
@@ -21,7 +21,7 @@
     /usr/lib/@{multiarch}/libseccomp.so* mr,
     /lib/@{multiarch}/libseccomp.so* mr,
 
-    @LIBEXECDIR@/snap-confine r,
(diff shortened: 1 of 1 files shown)


jdstrand:
FYI, uploaded 1.0.42-0ubuntu2 to yakkety for this change. When uploading a new version to yakkety, please don't drop this patch.

jdstrand:
FYI, the testsuite failure is unrelated to this commit:

```
2016/09/28 21:06:42 Cannot allocate linode:ubuntu-16.04-64-grub: authentication failed
```

jdstrand:
FYI, there is no bug for this.

jdstrand:
Actually, LP: #1628616 would fit.

PR #160, opened 2016-09-28
allow snap-confine to mount on /dev/pts/ptmx for LXD with /dev/ptmx symlink

jdstrand:


jdstrand:
FYI, this was uploaded to yakkety in 1.0.42-0ubuntu3. If preparing a new upload, please don't drop this patch.

jdstrand:
The test failure has nothing to do with this PR:

```
2016/09/28 22:36:58 Cannot allocate linode:ubuntu-16.04-64-grub: authentication failed
```

zyga:
The patch looks good, is there a bug number tracking this?

jdstrand:
@zyga there is no bug.

jdstrand:
Actually, LP: #1628616 would fit.

PR #162, opened 2016-10-05
Set PATH unconditionally even if mount namespace is set up

zyga:
This patch corrects a bug where PATH would be correctly set only on the
first invocation of a given snap's application. After the first
invodation the mount namespace would be already initialized and the
relevant code that sets PATH would not run. Now this is corrected and
PATH is always set.

Fixes: https://bugs.launchpad.net/ubuntu/+source/snap-confine/+bug/1630479


chipaca (approved):

jdstrand:
+1 as is, however, it occurred to me to ask why we don't do this after dropping privileges (eg, just before execv()). Don't block on this but feel free to move later if you want.

PR #163, opened 2016-10-05
Disable quirks on all-snap systems

zyga:
The quirks code makes sense only on a classic system. In an all-snap
world it prevents access to essential directories in the real /var/lib
directory (such as extrausers)

Fixes: https://bugs.launchpad.net/snap-confine/+bug/1630492


mvo5:
:+1:

zyga:
I just verified this on an all-snap system. Merging.

PR #167, opened 2016-10-06
drop 'owner' check on mountinfo and allow write to @{PROC}/[0-9]*/attr/current

jdstrand:
Due to a kernel bug, the ouid is not being set correctly for /proc accesses by
setuid processes running in user namespaces. While the kernel needs to be
fixed, drop the 'owner' match on @{PROC}/*/mountinfo for now.

Since we are using aa_change_hat(), snap-confine needs to be able to write to
@{PROC}/[0-9]*/attr/current.

Bug: https://launchpad.net/bugs/1630789

### DIFF
--- a/src/snap-confine.apparmor.in
+++ b/src/snap-confine.apparmor.in
@@ -208,13 +208,16 @@
     /run/snapd/ns/*.lock rwk,
     /run/snapd/ns/*.mnt rw,
     ptrace (tracedby) peer=@LIBEXECDIR@/snap-confine//mount-namespace-capture-helper,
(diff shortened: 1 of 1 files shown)


jdstrand:
Fyi, the test is failing for an unrelated reason: `2016/10/06 14:14:03 Cannot allocate linode:ubuntu-16.04-64-grub: authentication failed`

tyhicks:
Looks good to me. Thanks!

jdstrand:
This has been uploaded to Ubuntu 16.10.

PR #169, opened 2016-10-11
Add spread tests for mount namespace layout

zyga:
This patch adds a simple test that looks at /proc/self/mountinfo within
confined applications. The test discards some information that is too
variable to be useful but retains the most essential facts.

This test will be changed by the upcoming /media sharing patches. It is
intended to be useful for before/after comparison.


spread-tests/main/mount-ns-layout/process.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+import sys
+import json
+import re
+
+_boring_fs = set(['cgroup', 'fusectl', 'debugfs', 'pstore', 'securityfs', 'mqueue', 'hugetlbfs'])
```
tyhicks: Why are we ignoring these filesystems?

At least cgroup, debugfs, and securityfs are very security sensitive. At least a comment on the definition of a boring filesystem would be helpful.

spread-tests/main/mount-ns-layout/expected.json
```diff
@@ -0,0 +1,662 @@
...
+    "opt_fields": [
+      "master:renumbered/3"
+    ],
+    "root_dir": "/"
+  },
+  {
+    "fs_type": "ext4",
+    "mount_opts": "rw,noatime",
+    "mount_point": "/etc",
+    "mount_src": "/dev/sda",
```
tyhicks: Isn't the `/dev/sda` portion going to be brittle? For example, if testing in QEMU using a virtio, I suspect this will be `/dev/vda`.

tyhicks (commented):
This looks like a really nice test. I didn't spend too much time looking at the mountinfo contents and focused mostly on what the test was verifying at a higher level. I only had two comments/questions that I'd like cleared up before giving my ack.

spread-tests/main/mount-ns-layout/process.py
zyga: Ha, because I wanted to make it shorter (because there was an instability that I joust found and fixed). I'll amend this test to ignore nothing.

spread-tests/main/mount-ns-layout/expected.json
zyga: Nice catch, I'll change it to `/dev/BLOCK` or something like this

zyga:
I fixed both comments and more (the processing script was lousy so I wrote a real one)

zyga:
FYI: I finally fixed the non-determinism. It appears that there are differences between the core and ubuntu-core snaps. This is now reflected in the test.

PR #173, opened 2016-10-20
Make /writable unbindable

zyga:
This patch fixes lp:1635194 by making /writable unbindable so that it
doesn't explode across the mount namespaces of individual snaps.


zyga:
NB: Please don't merge this until we understand the problem better.

zyga:
I have a test that I want to land to snapd that measures number of mount table entries as subsequent snaps are executed. Let's postpone this until it lands and gives us some insight.

PR #174, opened 2016-10-25
Define MS_LAZYTIME if not known

vosst:


zyga (approved):
Thanks, I had this fixed locally but didn't push it yet.

The value is correct and checks out.

PR #177, opened 2016-10-26
system-shutdown helper

chipaca:


src/system-shutdown.c
```diff
@@ -0,0 +1,153 @@
+#include <stdbool.h>            // bools
```
zyga: Add the header please

src/system-shutdown.c
```diff
@@ -0,0 +1,153 @@
...
+
+                        if (strcmp("/proc", mount_dir) == 0) {
+                                continue;
+                        }
+
+                        if (strstr(mount_dir, "writable")) {
+                                had_writable = true;
+                        }
+
+                        if (umount(mount_dir) == 0) {
```
zyga: It would be good to document the logic below.

src/system-shutdown.c
```diff
@@ -0,0 +1,153 @@
...
+                                int fd = open(mount_src, O_RDONLY);
+                                if (fd >=0) {
```
zyga: Run "make fmt" on the tree please.

src/system-shutdown.c
```diff
@@ -0,0 +1,153 @@
...
+                                        ioctl(fd, LOOP_CLR_FD);
```
zyga: I'd rather we only run the ioctl if we know we're supposed to and then that we check for error codes.

src/system-shutdown.c
```diff
@@ -0,0 +1,153 @@
...
+                        if (strcmp("/dev", mount_dir) == 0) {
+                                continue;
+                        }
+
+                        if (strcmp("/proc", mount_dir) == 0) {
+                                continue;
+                        }
+
+                        if (strstr(mount_dir, "writable")) {
+                                had_writable = true;
```
zyga: This will trigger on a snap called `writable`. The test should be more robust IMO.

src/system-shutdown.c
```diff
@@ -0,0 +1,153 @@
...
+int main(int argc, char *argv[]) {
+        errno = 0;
+        if (getpid() != 1) {
+                fprintf(stderr, "This is a shutdown helper program; don't call it directly.\n");
+                exit(1);
+        }
+
+        kmsg("started.");
+
+        if (mkdir("/writable", 0755) < 0) {
```
zyga: Can you document where this program is expecting to execute and what happens if /writable already exists?

src/system-shutdown.c
```diff
@@ -0,0 +1,153 @@
...
+                die("unable to mkdir");
```
zyga: `"Cannot create directory /writable"`

zyga (changes requested):
Looks good as as a first stab at this problem but we should make it a little bit more robust.

src/system-shutdown.c
chipaca: only if the snap survived into this environment though, but point taken.

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
 
+static void detach_loop(const char *src)
+{
+	int fd = open(src, O_RDONLY);
+	if (fd < 0) {
+		kmsg("* unable to open loop device %s: %s", src,
+		     strerror(errno));
+	} else {
+		if (ioctl(fd, LOOP_CLR_FD) < 0) {
+			kmsg("* unable to disassociate loop device %ss: %s",
```
zyga, line 106: You might use `%m`(glibc extension) for the `strerr` conversion.

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
+			const char *src = mountinfo_entry_mount_source(cur);
+			unsigned major = mountinfo_entry_dev_major(cur);
+
+			cur = next_mountinfo_entry(cur);
+
+			if (streq("/", dir)) {
+				continue;
+			}
+
+			if (streq("/dev", dir)) {
```
zyga, line 141: Can you please document why this is not unmounted.

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
+		had_writable = false;
+		did_umount = false;
+		while (cur) {
+			const char *dir = mountinfo_entry_mount_dir(cur);
+			const char *src = mountinfo_entry_mount_source(cur);
+			unsigned major = mountinfo_entry_dev_major(cur);
+
+			cur = next_mountinfo_entry(cur);
+
+			if (streq("/", dir)) {
```
zyga, line 137: Can you please document why this is not unmounted?

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
+				continue;
+			}
+
+			if (streq("/dev", dir)) {
+				continue;
+			}
+
+			if (streq("/proc", dir)) {
```
zyga, line 145: I assume it is because of /proc/self/mountinfo but it would be good to document why this is not unmounted.

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
+				continue;
+			}
+
+			if (major != 0 && major != 7
```
zyga, line 149: Can we have a non-magic constant here please. I assume you want `LOOP_MAJOR` here (from `#include <linux/major.h>`

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
+			    && endswith(dir, "/writable")) {
+				had_writable = true;
+			}
+
+			if (umount(dir) == 0) {
+				if (major == 7) {
```
zyga, line 155: Ditto `LOOP_MAJOR`

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
+
+	/*
+
+	   This program is started by systemd exec'ing the "shutdown" binary
+	   inside what used to be /run/initramfs. That is: the system's
+	   /run/initramfs is now /, and the old / is now /oldroot. Our job is
+	   to disentagle /oldroot and /oldroot/writable, which contain each
+	   other in the "live" system. We do this by creating a new /writable
+	   and moving the old mount there, previous to which we need to unmount
+	   as much as we can. Having done that we should be able to detah the
```
zyga, line 187: Typo, you want `detach`

src/system-shutdown.c
```diff
@@ -1,153 +1,235 @@
...
+	if (umount_all()) {
+		kmsg("- found no hard-to-unmount writable partition.");
+	} else {
+		if (mount("/oldroot/writable", "/writable", NULL, MS_MOVE, NULL)
+		    < 0) {
+			die("cannot move writable out of the way");
+		}
+
+		bool ok = umount_all();
+		kmsg("%c was %s to unmount writable cleanly", ok ? '-' : '*',
```
zyga, line 205: What is the semantics of `-`and `*` in kmsg?

zyga (changes requested):
Looks much better, thank you. I've added a few more notes. As mentioned on telegram we should also try to add a few simple tests for trivial functions like `streq`or `endswith`

src/system-shutdown.c
chipaca: I looked in `loop(4)` and `MAKEDEV(8)` looking for exactly that include, thanks!

PR #178, opened 2016-10-27
Add snapd specific system-shutdown executable

zyga:
This branch picks up #177 which adds the `system-shutdown` helper executable. This executable is designed for shutting down core (aka all-snap) systems correctly.

There are still a few more things to do but this pull requests adds some unit tests and tweaks the code slightly.


jdstrand:
Should an ixr rule for this new binary be added to the shutdown interface? Will systemd's shutdown via DBus API call this helper?

zyga:
@jdstrand AFAIK the way this works is that on shutdown, there's a specific directory in `/run` that you can populate and then systemd will call it by default. I don't think we need to change any of the profiles unless systemd itself becomes confined.

PR #180, opened 2016-10-31
Tweak error messages for some consistency

zyga:
This patch makes a few ancient error messages more consistent with the
"cannot ..." scheme we have everywhere else. A few things are still more
awkward than they need to be but those require code simplifications that
should be done separately.


zyga (commented):
Self-review, +1

PR #183, opened 2016-11-18
Detach the hostfs version of /sys

zyga:
This patch detaches (aka umount --lazy) or umount2(1) with MNT_DETACH
flag the second sysfs that is visible from /var/lib/snapd/hostfs/sys
after the pivot_root call.

This is done so that the hostfs is more empty (second sysfs is not
really needed as we can always re-associated with pid-1 ns to get back
to the outer view) and so that certain application (*cough* docker
*cough*) that actually scan the mount table for sysfs don't find the
wrong one. The "wrong" one works exactly the same as /sys but is denied
by apparmor profiles.


lool:
Hi,
Zygmunt told me how to enter the target mount namespace:
sudo nsenter -m/run/snapd/ns/docker.mnt

I can confirm that if I "umount --lazy /var/lib/snapd/hostfs/sys" in there, it fixes subsequent docker runs.

NB: Zygmunt considers removing /proc and /dev too, and I can confirm that lazy unmounting /var/lib/snapd/hostfs/dev and /var/lib/snapd/hostfs/proc works there too, and docker keeps working.

Cheers,
- Loïc

src/mount-support.c
```diff
@@ -495,6 +495,16 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
 		die("cannot perform operation: mount --make-rslave %s",
 		    SC_HOSTFS_DIR);
 	}
+	// Detach the hostfs version of sysfs. Some software may want to inspect
+	// the mount table to look for a particular filesystem. If that said
+	// software unlucklily picks the hostfs "version" of. The prominent example
```
jdstrand: Typo: 'unlucklily'. This comment is phrased weird. How about:

```
// Detach the redundant hostfs version of sysfs since it shows up in the mount
// table and software inspecting the mount table may become confused (eg,
// docker and LP: #162601).
```

jdstrand (approved):

src/mount-support.c
zyga: Thanks! Indeed much shorter and to the point

PR #186, opened 2016-11-21
Detach the hostfs version of /proc

zyga:
This patch detaches (aka umount --lazy) or umount2(1) with MNT_DETACH
flag the second /proc that is visible from /var/lib/snapd/hostfs/proc
after the pivot_root call.


jdstrand:
LGTM

PR #187, opened 2016-11-21
mkdir XDG_RUNTIME_DIR (LP: #1620442)

jdstrand:


jdstrand:
The travis spread tests are failing due to authentication issues. The spread test I added passes but other unrelated tests are failing due to recent commits.

src/user-support.c
```diff
@@ -102,10 +102,32 @@ void setup_user_data()
...
+
+	if (xdg_runtime_dir == NULL)
+		return;
+	// Only support absolute paths.
+	if (xdg_runtime_dir[0] != '/') {
+		die("XDG_RUNTIME_DIR must be an absolute path");
+	}
+
+	errno = 0;
+	mkpath(xdg_runtime_dir);
```
zyga, line 126: I'll pick this up shortly so don't worry but I have two comments:

 - the error message from mkpath is confusing (it was designed for something else)
 - don't we need any apparmor changes for this to pass?

I'll fork and iterate to make sure this lands.

jdstrand: I made the apparmor changes as part of this PR. See src/snap-confine.apparmor.in.

Thanks for the review!

PR #188, opened 2016-11-23
Add apparmor support module

zyga:
This patch adds a new set of files, mimicking the xxx-support pattern
used for other changes, that so some extent abstracts the use of
apparmor in snap-confine.

Even when snap-confine is not compiled with apparmor support the same
APIs are available and gracefully degrade to no-ops.

As a small extension, snap confine can now know if it is confined (e.g.
a development version running in a random directory is not confined) and
will no longer crash when changing hats via aa_change_hat().

The patch doesn't yet change any of the tree to use the new functions.
This will be done in the next commit.

### DIFF
--- a/src/apparmor-support.c
+++ b/src/apparmor-support.c
@@ -1,0 +1,112 @@
+/*
+ * Copyright (C) 2016 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "apparmor-support.h"
+
+#include <stdbool.h>
+#include <string.h>
+#ifdef HAVE_APPARMOR
+#include <sys/apparmor.h>
+#endif				// ifdef HAVE_APPARMOR
+
+#include "cleanup-funcs.h"
+#include "utils.h"
+
+// NOTE: Those constants map exactly what apparmor is returning and cannot be
+// changed without breaking apparmor functionality.
+#define SC_AA_ENFORCE_STR "enforce"
+#define SC_AA_COMPLAIN_STR "complain"
+
+enum sc_mode {
+	// The enforcement mode was not recognized.
+	SC_AA_INVALID = -1,
+	// The enforcement mode is not applicable because apparmor is disabled.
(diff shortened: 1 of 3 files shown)


chipaca (approved):

src/apparmor-support.h
```diff
@@ -0,0 +1,91 @@
...
+enum sc_apparmor_mode {
+	// The enforcement mode was not recognized.
+	SC_AA_INVALID = -1,
+	// The enforcement mode is not applicable because apparmor is disabled.
+	SC_AA_NOT_APPLICABLE = 0,
+	// The enforcement mode is "enforcing"
+	SC_AA_ENFORCE = 1,
+	// The enforcement mode is "complain"
+	SC_AA_COMPLAIN,
+};
```
tyhicks, line 37: There is a new'ish mode (as of the 16.04 kernel), called "mixed". It means that two or more AppArmor profiles are stacked together but not all of their enforcement modes are the same. The mode string returned from aa_getcon() is "mixed".

src/apparmor-support.c
```diff
@@ -0,0 +1,92 @@
...
+#define SC_AA_UNCONFINED_STR "unconfined"
+
+void sc_init_apparmor_support(struct sc_apparmor *apparmor)
+{
+#ifdef HAVE_APPARMOR
+	char *label __attribute__ ((cleanup(sc_cleanup_string))) = NULL;
+	char *mode = NULL;	// mode cannot be free'd
+	if (aa_getcon(&label, &mode) < 0) {
+		die("cannot query current apparmor profile");
+	}
```
tyhicks, line 80: This misses the situation where AppArmor support is compiled into snap-confine but AppArmor has been explicitly disabled by passing "apparmor=0" on the kernel command line. aa_getcon() will return -1 with errno set to EINVAL in this condition but that errno unfortunately overlaps with some other conditions.

The definitive way to see if AppArmor is enabled is aa_is_enabled(). You should call it first and, if it returns 1, proceed to calling aa_getcon() to check if snap-confine is confined. See the aa_is_enabled() man page for details.

src/apparmor-support.c
```diff
@@ -0,0 +1,92 @@
...
+		apparmor->is_confined = false;
+	}
+	// Look at mode, it must be one of the well known strings.
+	if (mode != NULL && strcmp(mode, SC_AA_COMPLAIN_STR) == 0) {
+		apparmor->mode = SC_AA_COMPLAIN;
+	} else if (mode != NULL && strcmp(mode, SC_AA_ENFORCE_STR) == 0) {
+		apparmor->mode = SC_AA_ENFORCE;
+	} else {
+		apparmor->mode = SC_AA_INVALID;
+	}
```
tyhicks, line 99: You'll want to check for "mixed" mode here. I've left a more descriptive review comment in the accompanying header file.

tyhicks (changes requested):
This looks good. I only have two small suggestions that I left inline.

src/apparmor-support.c
zyga: +1 Thanks for this feedback. I'll do this.

src/apparmor-support.h
zyga: Thanks, I'll update the code to handle this as well.

src/apparmor-support.c
zyga: Ack

PR #194, opened 2016-11-24
Make messages from die() nicer

zyga:


chipaca (approved):
This is OK. I think getting a copy of errno is unnecessary in this context, but it can't hurt.

PR #199, opened 2016-12-01
Add error support code

zyga:
This patch adds die()-like APIs that can carry error information around
the control flow. They are meant to provide richer error information  in
cases where the error is not fatal and needs to be returned to the
caller. The APIs are 100% tested and also integrate with die() where
appropriate().

### DIFF
--- a/src/error.h
+++ b/src/error.h
@@ -68,8 +68,8 @@ struct sc_error *sc_error_init_from_errno(int errno_copy, const char *msgfmt,
 /**
  * Get the error domain out of an error object.
  *
- * The error domain acts as a namespace for error codes. The return value must
- * not be released in any way.
+ * The error domain acts as a namespace for error codes.
+ * No change of ownership takes place.
  **/
 __attribute__ ((warn_unused_result, nonnull, returns_nonnull))
 const char *sc_error_domain(struct sc_error *err);
@@ -91,6 +91,7 @@ int sc_error_code(struct sc_error *err);
  * Get the error message out of an error object.
  *
  * The error message is bound to the life-cycle of the error object.
+ * No change of ownership takes place.
  **/
 __attribute__ ((warn_unused_result, nonnull, returns_nonnull))
 const char *sc_error_msg(struct sc_error *err);
@@ -128,9 +129,11 @@ void sc_die_on_error(struct sc_error *error);
  * This tries to forward an error to the caller. If this is impossible because
  * the caller did not provide a location for the error to be stored then the
  * sc_die_on_error() is called as a safety measure.
+ *
+ * Change of ownership takes place and the error is now stored in the recipient.
  **/
 __attribute__ ((nonnull(1)))
-void sc_error_forward(struct sc_error **recepient, struct sc_error *error);
+void sc_error_forward(struct sc_error **recipient, struct sc_error *error);
 
 /**
  * Check if a given error matches the specified domain and code.
--- a/src/error.h
+++ b/src/error.h
@@ -49,7 +49,7 @@ struct sc_error;
  *
  * This function calls die() in case of memory allocation failure.
  **/
-__attribute__ ((format(printf, 3, 4)))
+__attribute__ ((warn_unused_result, format(printf, 3, 4), returns_nonnull))
 struct sc_error *sc_error_init(const char *domain, int code, const char *msgfmt,
 			       ...);
 
@@ -61,7 +61,7 @@ struct sc_error *sc_error_init(const char *domain, int code, const char *msgfmt,
  *
  * This function calls die() in case of memory allocation failure.
  **/
-__attribute__ ((format(printf, 2, 3)))
+__attribute__ ((warn_unused_result, format(printf, 2, 3), returns_nonnull))
 struct sc_error *sc_error_init_from_errno(int errno_copy, const char *msgfmt,
 					  ...);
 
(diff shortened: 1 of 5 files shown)


src/error.h
```diff
@@ -0,0 +1,128 @@
...
+ * This function calls die() in case of memory allocation failure.
+ **/
+__attribute__ ((format(printf, 2, 3)))
+struct sc_error *sc_error_init_from_errno(int errno_copy, const char *msgfmt,
+					  ...);
+
+/**
+ * Get the error domain out of an error object.
+ *
+ * The error domain acts as a namespace for error codes. The return value must
```
vosst: nit: I would propose to rephrase the doc string a little bit: "No change of ownership takes place".

src/error.h
```diff
@@ -0,0 +1,128 @@
...
+ * code is reserved for this error and it's not something that the programmer
+ * can rely on programmatically. This can be used to return an error message
+ * without having to allocate a distinct code for each one.
+ **/
+int sc_error_code(struct sc_error *err);
+
+/**
+ * Get the error message out of an error object.
+ *
+ * The error message is bound to the life-cycle of the error object.
```
vosst, line 93: See previous comment on ownership.

src/error.h
```diff
@@ -0,0 +1,128 @@
...
+/**
+ *
+ * Die if there's an error.
+ *
+ * This function is a correct way to die() if the passed error is not NULL.
+ *
+ * The error message is derived from the data in the error, using the special
+ * errno domain to provide additional information if that is available.
+ **/
+void sc_error_die(struct sc_error *error);
```
vosst: Maybe rename to sc_error_die_if.

src/error.h
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * The domain is a cookie-like string that allows the caller to distinguish
+ * between "namespaces" of error codes. It should be a static string that is
+ * provided by the caller. Both the domain and the error code can be retrieved
+ * later.
+ *
+ * This function calls die() in case of memory allocation failure.
+ **/
+__attribute__ ((format(printf, 3, 4)))
+struct sc_error *sc_error_init(const char *domain, int code, const char *msgfmt,
```
vosst, line 53: nit: _new or _alloc might go better with _free.

src/error.h
```diff
@@ -0,0 +1,128 @@
...
+void sc_error_die(struct sc_error *error);
+
+/**
+ * Forward an error to the caller.
+ *
+ * This tries to forward an error to the caller. If this is impossible because
+ * the caller did not provide a location for the error to be stored then the
+ * sc_error_die() is called as a safety measure.
+ **/
+void sc_error_forward(struct sc_error **recepient, struct sc_error *error);
```
vosst: I'm slightly confused by this method (likely because I lack a bit of context): I would have expected sc_error* sc_error_forward_or_die(struct sc_error* error);, with the function dying if error is NULL.

vosst (commented):
Code is looking good overall. I left a couple of comments inline, mostly nits.

src/error.h
zyga: This is essentially a way to enforce a contract that an error is forwarded to the caller or we die right here, right now. For usage context look at https://github.com/snapcore/snap-confine/blob/arguments/src/snap-confine-args.c#L135

zyga:
Thanks for the review. I'll fix something in another branch and return to this quickly.

src/error.h
chipaca: also, consider attributes `nonnull` (unlless `format` wanrs about a null `msgfmt` already), `warn_unused_result`, `returns_nonnull`.

src/error.h
chipaca: here a `nonnull` for the recepient would be good i think

src/error.h
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Initialize an errno-based error.
+ *
+ * The error carries a copy of errno and a custom error message as designed by
+ * the caller. See sc_error_init() for a more complete description.
+ *
+ * This function calls die() in case of memory allocation failure.
+ **/
+__attribute__ ((format(printf, 2, 3)))
```
chipaca: maybe some of those apply here as well

src/error.h
```diff
@@ -0,0 +1,128 @@
...
+struct sc_error *sc_error_init_from_errno(int errno_copy, const char *msgfmt,
+					  ...);
+
+/**
+ * Get the error domain out of an error object.
+ *
+ * The error domain acts as a namespace for error codes. The return value must
+ * not be released in any way.
+ **/
+const char *sc_error_domain(struct sc_error *err);
```
chipaca, line 75: and all of these could probably benefit from `nonnull`

src/error.h
chipaca: or `sc_error_die_maybe` |-)

chipaca (commented):
It looks good. I've pointed out some attributes you could use to helper the compiler errors/warnings, but those are pico-nits.

One thing that does bother me a little is that you pay for formatting things up front instead of at die time. Had you considered carrying around the `msgfmt` and `va_list` and doing the `vasprintf` at the end if needed?

zyga:
@chipaca In C that's generally hard as va_list is hard to "store" for longer than a function call and even if you could then there's no guarantee that value stored there are still sane (e.g. pointers). One of the non-perks of manual memory management :/

src/error.h
zyga: I called this `sc_die_on_error` to make it dead obvious the cost of the naming scheme.

src/error.h
zyga: Changed, thanks.

src/error.h
zyga: Hmm, this feels C++-ish but as long as we're consistent that's okay. Changed

vosst:
LGTM.

tyhicks:
I took a quick look at this PR and it looks ok to me.

PR #200, opened 2016-12-05
Add module for parsing command line arguments.

zyga:
This patch adds a new module that does what current snap-confine main() does
with regards to argument parsing. The module is fully tested and has simple API
sufficient to drive rest of snap-confine.


zyga:
This was picked up https://github.com/snapcore/snapd/pull/2416

